PR #2, opened 2018-11-14
Fix compiler warning (change CHAR8 to unsigned char)

neilsh-msft:
- Fix compiler warning on ARM64, d2i_PKCS7 is passed (const CHAR8 **) but expects (const unsigned char **).  Brings this usage of d2i_PKCS7 inline with the usage in other files.

### DIFF
--- a/CryptoPkg/Library/BaseCryptLib/Pk/CryptPkcs7VerifyEku.c
+++ b/CryptoPkg/Library/BaseCryptLib/Pk/CryptPkcs7VerifyEku.c
@@ -451,7 +451,7 @@ VerifyEKUsInPkcs7Signature (
   //
   // Create the PKCS7 object.
   //
-  Pkcs7 = d2i_PKCS7(NULL, (CONST CHAR8 **)&Temp, (INT32)SignedDataSize);
+  Pkcs7 = d2i_PKCS7(NULL, (const unsigned char **) &Temp, (INT32)SignedDataSize);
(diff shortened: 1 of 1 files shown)


CryptoPkg/Library/BaseCryptLib/Pk/CryptPkcs7VerifyEku.c
```diff
@@ -451,7 +451,7 @@ VerifyEKUsInPkcs7Signature (
   //
   // Create the PKCS7 object.
   //
-  Pkcs7 = d2i_PKCS7(NULL, (CONST CHAR8 **)&Temp, (INT32)SignedDataSize);
+  Pkcs7 = d2i_PKCS7(NULL, (const unsigned char **) &Temp, (INT32)SignedDataSize);
```
SurfaceDMFUser1, line 454: This should be const unsigned char **

SurfaceDMFUser1 (approved):
Looks good Neil.

CryptoPkg/Library/BaseCryptLib/Pk/CryptPkcs7VerifyEku.c
SurfaceDMFUser1, line 454: Yes, it should be (const unsigned char**) or to be even more UEFI like (const UINT8**) or event better we could take the casting out completely because Temp is declared as UINT8*.

PR #13, opened 2020-05-09
Bring in FmpDependencyLib and FmpDependencyCheckLib from EDKII

kuqin12:
Bring in FmpDependencyLib and FmpDependencyCheckLib from EDKII:
1. Verified on a real system with FmpDependencyCheckLibNull;
2. Verified on a real system with FmpDevicePkg\Test\UnitTest\Library\FmpDependencyLib\FmpDependencyLibUnitTestsUefi.inf;
3. Updated token number for gFmpDevicePkgTokenSpaceGuid.PcdFmpDxeRequiredEKU as it conflicted with newly added Pcd from EDKII;
4. Built a capsule with real external dependency on other firmware and tested on UEFI image with FmpDependencyCheckLib. FmpDxe driver properly logged unsatisfied dependency. OS also triggered a retry after any other successful capsule update;


FmpDevicePkg/Test/UnitTest/Library/FmpDependencyLib/EvaluateDependencyUnitTest.c
```diff
@@ -0,0 +1,271 @@
...
+                     );
+
+  UT_ASSERT_EQUAL (EvaluationResult, TestContext->ExpectedResult);
+
+  return UNIT_TEST_PASSED;
+}
+
+/**
+  Initialize the unit test framework, suite, and unit tests for the
+  Base64 conversion APIs of BaseLib and run the unit tests.
```
spbrogan, line 182: comments need to be adjusted as this looks copy/paste from Base64 test

FmpDevicePkg/Test/UnitTest/Library/FmpDependencyLib/EvaluateDependencyUnitTest.c
```diff
@@ -0,0 +1,271 @@
...
+      goto EXIT;
+  }
+
+  //
+  // Populate the Unit Test Suite.
+  //
+  Status = CreateUnitTestSuite (&DepexEvalTests, Fw, "Evaluate Dependency Test", "FmpDependencyLib.EvaluateDependency", NULL, NULL);
+  if (EFI_ERROR (Status)) {
+    DEBUG ((DEBUG_ERROR, "Failed in CreateUnitTestSuite for DepexEvalTests\n"));
+    Status = EFI_OUT_OF_RESOURCES;
```
spbrogan, line 218: why hide the real status returned from CreateUnitTestSuite?  It is clearly an error so seems better to keep the return error than to mask it

FmpDevicePkg/Test/UnitTest/Library/FmpDependencyLib/EvaluateDependencyUnitTest.c
```diff
@@ -0,0 +1,271 @@
...
+    goto EXIT;
+  }
+
+  AddTestCase (DepexEvalTests, "Evaluate to True - 1", "Test1", EvaluateDependencyTest, NULL, NULL, &mBasicTestTrue1);
```
spbrogan, line 222: I would like better descriptions.  
From the name i would have expected an expression that was just hardcoded to TRUE. 
After looking over the test this is actually "(Version(A) > 0x00000001) && (Version(B) >= 0x00000003)"

So this test is actually testing greater than, two conditionals, and greater than or equal. 

I would suggest a simple test for each conditional type.  Then I would add two or more conditionals.  

But what is great about Unit tests is that they are easy and fast to confirm functionality defined in specification.  If you have a mapping that is closer to 1:1 then very easy to see that and very easy for someone to extend to add an edge case that fails.  For example i would test ">=" by doing three cases.  1 less, 1 greater, and 1 equal.

FmpDevicePkg/Library/FmpDependencyLib/FmpDependencyLib.c
```diff
@@ -472,18 +494,18 @@ ValidateImageDepex (
...
-  @param[in]   Dependencies    Pointer to the EFI_FIRMWARE_IMAGE_DEP.
+  @param[in]  Dependencies   Pointer to the EFI_FIRMWARE_IMAGE_DEP.
 
   @retval  The size of dependencies.
 
 **/
 UINTN
-GetDepexSize (
+EFIAPI
+GetDependencySize (
```
spbrogan, line 508: Should this have a maxsize parameter?  This looks like a public API so it probably needs to do checks.  The while loop could be endless as written.

FmpDevicePkg/Library/FmpDependencyCheckLibNull/FmpDependencyCheckLibNull.uni
```diff
@@ -0,0 +1,13 @@
+// /** @file
+// Provides FMP capsule dependency check services when updating the firmware
+// image of a FMP device.
```
spbrogan, line 3: What is the point of this file if it doesn't indicate that this is a NULL (do nothing) instance?

FmpDevicePkg/Library/FmpDependencyCheckLibNull/FmpDependencyCheckLibNull.inf
```diff
@@ -0,0 +1,30 @@
+## @file
+#  Provides FMP capsule dependency check services when updating the firmware
+#  image of a FMP device.
```
spbrogan, line 3: I would suggest comment about being a NULL/ do nothing instance

FmpDevicePkg/Include/Library/FmpDependencyCheckLib.h
```diff
@@ -0,0 +1,71 @@
...
+  @param[in]  DepexSize   Size, in bytes, of the Fmp dependency.
+
+  @retval  EFI_SUCCESS       Save Fmp dependency succeeds.
+  @retval  EFI_UNSUPPORTED   Save Fmp dependency is not supported.
+  @retval  Others            Save Fmp dependency fails.
+
+**/
+EFI_STATUS
+EFIAPI
+SaveFmpDependency (
```
spbrogan, line 51: This API seems like it would be better in FmpDeviceLib

FmpDevicePkg/Include/Library/FmpDependencyCheckLib.h
```diff
@@ -0,0 +1,71 @@
...
+
+  @param[out]  DepexSize   Size, in bytes, of the dependency.
+
+  @retval  The pointer to dependency.
+  @retval  NULL
+
+**/
+EFI_FIRMWARE_IMAGE_DEP*
+EFIAPI
+GetFmpDependency (
```
spbrogan, line 67: This API seems like it would better in FmpDeviceLib

FmpDevicePkg/Library/FmpDependencyCheckLib/FmpDependencyCheckLib.c
```diff
@@ -0,0 +1,533 @@
...
+  UINT32                            PackageVersion;
+  CHAR16                            *PackageVersionName;
+  UINTN                             DepexSize;
+  UINTN                             NumberOfFmpInstance;
+  EFI_FIRMWARE_IMAGE_DESCRIPTOR     **FmpImageInfoBuf;
+  FMP_DEPEX_CHECK_VERSION_DATA      *FmpVersions;
+  UINTN                             FmpVersionsCount;
+  BOOLEAN                           IsSatisfied;
+
+  if (!FeaturePcdGet (PcdFmpDependencyCheckEnable)) {
```
spbrogan, line 113: I am interested to understand why this PCD is needed.  If you want to always return true just use the null instance.  See here 
https://github.com/microsoft/mu_tiano_plus/pull/13/files?file-filters%5B%5D=.c#diff-c8927b0653c75167d759615ac36668deR27

FmpDevicePkg/Library/FmpDependencyCheckLib/FmpDependencyCheckLib.c
```diff
@@ -0,0 +1,533 @@
...
+             (VOID **) &Depex,
+             &Size
+             );
+  if (EFI_ERROR (Status)) {
+    if (Status != EFI_NOT_FOUND) {
+      //
+      // Treat EFI_NOT_FOUND as no dependency.
+      // Other errors are treated as dependency evaluates to FALSE.
+      //
+      Size = 2;
```
spbrogan, line 460: This is one of the cases where you will now have a device that is stuck and always evaluating to false?  Do you really want this denial of service?

FmpDevicePkg/Library/FmpDependencyCheckLib/FmpDependencyCheckLib.c
```diff
@@ -0,0 +1,533 @@
...
+
+  VariableLock = NULL;
+  Status = gBS->LocateProtocol (
+                  &gEdkiiVariableLockProtocolGuid,
+                  NULL,
+                  (VOID **)&VariableLock
+                  );
+  if (EFI_ERROR (Status) || VariableLock == NULL) {
+    DEBUG ((DEBUG_ERROR, "FmpDependencyCheckLib: Failed to locate Variable Lock Protocol (%r).\n", Status));
+    return EFI_SUCCESS;
```
spbrogan, line 508: so if you fail to lock...you return success?

spbrogan (changes requested):
I need to review more but this is enough to start talking.  

There are a lot of conditional choices and checks in this code.  I think this might be a case where code coverage would be helpful to see how much test coverage you are actually getting.  You can enable this for GCC5 builds by doing this. 

https://github.com/spbrogan/edk2-staging/commit/ff566ee0ff65240c93b8c83e79b2dd6b990a488e

FmpDevicePkg/Library/FmpDependencyCheckLib/FmpDependencyCheckLib.c
xuweiintel: If a library constructor returns error, it will block boot.

FmpDevicePkg/Library/FmpDependencyLib/FmpDependencyLib.c
xuweiintel: No maxsize parameter is a compromise. If the dependency comes from the EFI_FIRMWARE_IMAGE_DESCRIPTOR, we cannot know the maxsize from anywhere. This API is used in CheckFmpDependency() when checking the dependency from descriptor.
Your concern is correct.  Since the API ValidateDependency can output the depex size and have a maxsize parameter, so I will just remove this API and only use it privately in FmpDependencyCheckLib.

FmpDevicePkg/Library/FmpDependencyCheckLib/FmpDependencyCheckLib.c
xuweiintel: Yes. When errors occur, it means variable service is down, which indicates the entire system is unstable. We must deny current update and update system firmware firstly.

FmpDevicePkg/Include/Library/FmpDependencyCheckLib.h
xuweiintel: The advantage of this API in FmpDependencyCheckLib is that we can reduce the impacts to existing platforms, because there are too many FmpDeviceLib instances in a platform. Also, when platform implements the depex storage things, platform might choose same way for most FmpDevice. Finally, the depex saving and getting is a part of depex check. So, I think it is reasonable to be here in FmpDependencyCheckLib.

FmpDevicePkg/Include/Library/FmpDependencyCheckLib.h
xuweiintel: Same as above.

FmpDevicePkg/Test/UnitTest/Library/FmpDependencyLib/EvaluateDependencyUnitTest.c
xuweiintel: Thanks for finding this. I will update it.

FmpDevicePkg/Test/UnitTest/Library/FmpDependencyLib/EvaluateDependencyUnitTest.c
xuweiintel: Also copy/paste from Base64 test. I will update it.

FmpDevicePkg/Library/FmpDependencyCheckLibNull/FmpDependencyCheckLibNull.uni
xuweiintel: Thanks for finding this. I will update it.

FmpDevicePkg/Library/FmpDependencyCheckLibNull/FmpDependencyCheckLibNull.inf
xuweiintel: Thanks for finding this. I will update it.

FmpDevicePkg/Library/FmpDependencyCheckLib/FmpDependencyCheckLib.c
xuweiintel: It is intended to provide an option to disable dependency check when using the non-null instance. I don't have strong opinion on it, we can remove the PCD.

FmpDevicePkg/Library/FmpDependencyCheckLib/FmpDependencyCheckLib.c
spbrogan: is that worse than booting but failing to lock? When as a developer you think you are locked?

xuweiintel: I see. Thanks for pointing out this. I will update this to return actual error status.

spbrogan:
shared feedback with Intel.  this can close now.

PR #16, opened 2020-10-26
Drop the .pytool version of the readme

corthon:


corthon:
Dropping. This is obviated by the 2102 integration.

PR #51, opened 2022-11-08
Don't Set Access Attributes of Runtime MMIO Ranges

TaylorBeebe:
## Description

Passing in access attributes to SetMemorySpaceAttributes() will cause the existing attributes to be overwritten. The MMIO region should have the appropriate attributes applied during memory protection initialization and the attributes of the memory descriptor are inaccurate. Don't pass in any access attributes so the region stays secure.

## Breaking change?

No

## How This Was Tested

- Booting to the OS and running the paging audit app on Q35

## Integration Instructions

N/A

### DIFF
--- a/PrmPkg/PrmConfigDxe/PrmConfigDxe.c
+++ b/PrmPkg/PrmConfigDxe/PrmConfigDxe.c
@@ -155,7 +155,11 @@ SetRuntimeMemoryRangeAttributes (
     Status = gDS->SetMemorySpaceAttributes (
                     RuntimeMmioRanges->Range[Index].PhysicalBaseAddress,
                     (UINT64)RuntimeMmioRanges->Range[Index].Length,
-                    Descriptor.Attributes | EFI_MEMORY_RUNTIME
+                    // MU_CHANGE START: The memory space descriptor access attributes are not accurate. Don't pass
+                    //                  in access attributes so SetMemorySpaceAttributes() doesn't update them.
+                    // Descriptor.Attributes | EFI_MEMORY_RUNTIME
+                    EFI_MEMORY_RUNTIME
+                    // MU_CHANGE END
                     );
     ASSERT_EFI_ERROR (Status);
     if (EFI_ERROR (Status)) {


os-d (approved):

kenlautner (approved):

makubacki (approved):

makubacki:
Can you elaborate on "the attributes of the memory descriptor are inaccurate"? I don't recall what the attributes were coming into this function.

TaylorBeebe:
> Can you elaborate on "the attributes of the memory descriptor are inaccurate"? I don't recall what the attributes were coming into this function.

It grabs the existing memory space descriptor and ORs that with EFI_MEMORY_RUNTIME to pass it into SetMemorySpaceAttriubtes(). The SetMemorySpaceAttriubtes() function doesn't subsequently call gCpu->SetMemoryAttributes() if no CPU Architectural Attributes are passed in. Passing in EFI_MEMORY_RUNTIME just checks that the region is page aligned.

makubacki:
> > Can you elaborate on "the attributes of the memory descriptor are inaccurate"? I don't recall what the attributes were coming into this function.
> 
> It grabs the existing memory space descriptor and ORs that with EFI_MEMORY_RUNTIME to pass it into SetMemorySpaceAttriubtes(). The SetMemorySpaceAttriubtes() function doesn't subsequently call gCpu->SetMemoryAttributes() if no CPU Architectural Attributes are passed in. Passing in EFI_MEMORY_RUNTIME just checks that the region is page aligned.

Again, I don't recall all the details of this flow without looking further into the code, but `CoreConvertSpace()` in Gcd.c has the following comment:

```c
    // CPU arch attributes include page attributes and cache attributes.
    // Only page attributes supports to be cleared, but not cache attributes.
    // Caller is expected to use GetMemorySpaceDescriptor() to get the current
    // attributes, AND/OR attributes, and then calls SetMemorySpaceAttributes()
    // to set the new attributes.
    // So 0 CPU arch attributes should not happen as memory should always have
    // a cache attribute (no matter UC or WB, etc).
    //
    // Here, 0 CPU arch attributes will be filtered to be compatible with the
    // case that caller just calls SetMemorySpaceAttributes() with none CPU
    // arch attributes (for example, RUNTIME) as the purpose of the case is not
    // to clear CPU arch attributes.
```

Are you wanting to avoid the AND/OR operation described in the comment to prevent CPU arch attributes from being present to prevent `gCpu->SetMemoryAttributes()` from being called?

I understand runtime memory pages need to be page aligned and it checks that, but doesn't it do more than just check that? I thought this code would be executed as well taking a quick look at the file.

```
      //
      // Set attributes operation
      //
      case GCD_SET_ATTRIBUTES_MEMORY_OPERATION:
        if (CpuArchAttributes == 0) {
          //
          // Keep original CPU arch attributes when caller just calls
          // SetMemorySpaceAttributes() with none CPU arch attributes (for example, RUNTIME).
          //
          Attributes |= (Entry->Attributes & (EFI_CACHE_ATTRIBUTE_MASK | EFI_MEMORY_ATTRIBUTE_MASK));
        }

        Entry->Attributes = Attributes;
        break;
```

TaylorBeebe:
> > > Can you elaborate on "the attributes of the memory descriptor are inaccurate"? I don't recall what the attributes were coming into this function.
> > 
> > 
> > It grabs the existing memory space descriptor and ORs that with EFI_MEMORY_RUNTIME to pass it into SetMemorySpaceAttriubtes(). The SetMemorySpaceAttriubtes() function doesn't subsequently call gCpu->SetMemoryAttributes() if no CPU Architectural Attributes are passed in. Passing in EFI_MEMORY_RUNTIME just checks that the region is page aligned.
> 
> Again, I don't recall all the details of this flow without looking further into the code, but `CoreConvertSpace()` in Gcd.c has the following comment:
> 
> ```c
>     // CPU arch attributes include page attributes and cache attributes.
>     // Only page attributes supports to be cleared, but not cache attributes.
>     // Caller is expected to use GetMemorySpaceDescriptor() to get the current
>     // attributes, AND/OR attributes, and then calls SetMemorySpaceAttributes()
>     // to set the new attributes.
>     // So 0 CPU arch attributes should not happen as memory should always have
>     // a cache attribute (no matter UC or WB, etc).
>     //
>     // Here, 0 CPU arch attributes will be filtered to be compatible with the
>     // case that caller just calls SetMemorySpaceAttributes() with none CPU
>     // arch attributes (for example, RUNTIME) as the purpose of the case is not
>     // to clear CPU arch attributes.
> ```
> 
> Are you wanting to avoid the AND/OR operation described in the comment to prevent CPU arch attributes from being present to prevent `gCpu->SetMemoryAttributes()` from being called?
> 
> I understand runtime memory pages need to be page aligned and it checks that, but doesn't it do more than just check that? I thought this code would be executed as well taking a quick look at the file.
> 
> ```
>       //
>       // Set attributes operation
>       //
>       case GCD_SET_ATTRIBUTES_MEMORY_OPERATION:
>         if (CpuArchAttributes == 0) {
>           //
>           // Keep original CPU arch attributes when caller just calls
>           // SetMemorySpaceAttributes() with none CPU arch attributes (for example, RUNTIME).
>           //
>           Attributes |= (Entry->Attributes & (EFI_CACHE_ATTRIBUTE_MASK | EFI_MEMORY_ATTRIBUTE_MASK));
>         }
> 
>         Entry->Attributes = Attributes;
>         break;
> ```

My mistake - it will add the EFI_MEMORY_RUNTIME attribute.

I'm making this change because I want it to use the CPU attributes already present in the range
```
      case GCD_SET_ATTRIBUTES_MEMORY_OPERATION:
        if (CpuArchAttributes == 0) {
          //
          // Keep original CPU arch attributes when caller just calls
          // SetMemorySpaceAttributes() with none CPU arch attributes (for example, RUNTIME).
          //
          Attributes |= (Entry->Attributes & (EFI_CACHE_ATTRIBUTE_MASK | EFI_MEMORY_ATTRIBUTE_MASK));
        }

        Entry->Attributes = Attributes;
        break;
```

and NOT try to set the attributes to the CPU arch attributes passed in

```
  if (Operation == GCD_SET_ATTRIBUTES_MEMORY_OPERATION) {
    //
    // Call CPU Arch Protocol to attempt to set attributes on the range
    //
    CpuArchAttributes = ConverToCpuArchAttributes (Attributes);
    //
    // CPU arch attributes include page attributes and cache attributes.
    // Only page attributes supports to be cleared, but not cache attributes.
    // Caller is expected to use GetMemorySpaceDescriptor() to get the current
    // attributes, AND/OR attributes, and then calls SetMemorySpaceAttributes()
    // to set the new attributes.
    // So 0 CPU arch attributes should not happen as memory should always have
    // a cache attribute (no matter UC or WB, etc).
    //
    // Here, 0 CPU arch attributes will be filtered to be compatible with the
    // case that caller just calls SetMemorySpaceAttributes() with none CPU
    // arch attributes (for example, RUNTIME) as the purpose of the case is not
    // to clear CPU arch attributes.
    //
    if (CpuArchAttributes != 0) {
      if (gCpu == NULL) {
        Status = EFI_NOT_AVAILABLE_YET;
      } else {
        Status = gCpu->SetMemoryAttributes (
                         gCpu,
                         BaseAddress,
                         Length,
                         CpuArchAttributes
                         );
      }
```

The original call is doing exactly what should be done. However, in practice the attributes of the GCD memory space map are not currently accurate in Project Mu or EDK2, so passing in the attributes of the existing memory space descriptor won't work as intended.

makubacki:
> > > > Can you elaborate on "the attributes of the memory descriptor are inaccurate"? I don't recall what the attributes were coming into this function.
> > > 
> > > 
> > > It grabs the existing memory space descriptor and ORs that with EFI_MEMORY_RUNTIME to pass it into SetMemorySpaceAttriubtes(). The SetMemorySpaceAttriubtes() function doesn't subsequently call gCpu->SetMemoryAttributes() if no CPU Architectural Attributes are passed in. Passing in EFI_MEMORY_RUNTIME just checks that the region is page aligned.
> 
> My mistake - it will add the EFI_MEMORY_RUNTIME attribute.
> 
> I'm making this change because I want it to use the CPU attributes already present in the range
> 
> ```
>       case GCD_SET_ATTRIBUTES_MEMORY_OPERATION:
>         if (CpuArchAttributes == 0) {
>           //
>           // Keep original CPU arch attributes when caller just calls
>           // SetMemorySpaceAttributes() with none CPU arch attributes (for example, RUNTIME).
>           //
>           Attributes |= (Entry->Attributes & (EFI_CACHE_ATTRIBUTE_MASK | EFI_MEMORY_ATTRIBUTE_MASK));
>         }
> 
>         Entry->Attributes = Attributes;
>         break;
> ```
> 
> and NOT try to set the attributes to the CPU arch attributes passed in
> 
> ```
>   if (Operation == GCD_SET_ATTRIBUTES_MEMORY_OPERATION) {
>     //
>     // Call CPU Arch Protocol to attempt to set attributes on the range
>     //
>     CpuArchAttributes = ConverToCpuArchAttributes (Attributes);
>     //
>     // CPU arch attributes include page attributes and cache attributes.
>     // Only page attributes supports to be cleared, but not cache attributes.
>     // Caller is expected to use GetMemorySpaceDescriptor() to get the current
>     // attributes, AND/OR attributes, and then calls SetMemorySpaceAttributes()
>     // to set the new attributes.
>     // So 0 CPU arch attributes should not happen as memory should always have
>     // a cache attribute (no matter UC or WB, etc).
>     //
>     // Here, 0 CPU arch attributes will be filtered to be compatible with the
>     // case that caller just calls SetMemorySpaceAttributes() with none CPU
>     // arch attributes (for example, RUNTIME) as the purpose of the case is not
>     // to clear CPU arch attributes.
>     //
>     if (CpuArchAttributes != 0) {
>       if (gCpu == NULL) {
>         Status = EFI_NOT_AVAILABLE_YET;
>       } else {
>         Status = gCpu->SetMemoryAttributes (
>                          gCpu,
>                          BaseAddress,
>                          Length,
>                          CpuArchAttributes
>                          );
>       }
> ```
> 
> The original call is doing exactly what should be done. However, in practice the attributes of the GCD memory space map are not currently accurate in Project Mu or EDK2, so passing in the attributes of the existing memory space descriptor won't work as intended.

Okay, I can see that. Perhaps in Project Mu we should update that comment then with some of these additional details. Do you think this is a broader statement that should be made there?

TaylorBeebe:
> > > > > Can you elaborate on "the attributes of the memory descriptor are inaccurate"? I don't recall what the attributes were coming into this function.
> > > > 
> > > > 
> > > > It grabs the existing memory space descriptor and ORs that with EFI_MEMORY_RUNTIME to pass it into SetMemorySpaceAttriubtes(). The SetMemorySpaceAttriubtes() function doesn't subsequently call gCpu->SetMemoryAttributes() if no CPU Architectural Attributes are passed in. Passing in EFI_MEMORY_RUNTIME just checks that the region is page aligned.
> > 
> > 
> > My mistake - it will add the EFI_MEMORY_RUNTIME attribute.
> > I'm making this change because I want it to use the CPU attributes already present in the range
> > ```
> >       case GCD_SET_ATTRIBUTES_MEMORY_OPERATION:
> >         if (CpuArchAttributes == 0) {
> >           //
> >           // Keep original CPU arch attributes when caller just calls
> >           // SetMemorySpaceAttributes() with none CPU arch attributes (for example, RUNTIME).
> >           //
> >           Attributes |= (Entry->Attributes & (EFI_CACHE_ATTRIBUTE_MASK | EFI_MEMORY_ATTRIBUTE_MASK));
> >         }
> > 
> >         Entry->Attributes = Attributes;
> >         break;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and NOT try to set the attributes to the CPU arch attributes passed in
> > ```
> >   if (Operation == GCD_SET_ATTRIBUTES_MEMORY_OPERATION) {
> >     //
> >     // Call CPU Arch Protocol to attempt to set attributes on the range
> >     //
> >     CpuArchAttributes = ConverToCpuArchAttributes (Attributes);
> >     //
> >     // CPU arch attributes include page attributes and cache attributes.
> >     // Only page attributes supports to be cleared, but not cache attributes.
> >     // Caller is expected to use GetMemorySpaceDescriptor() to get the current
> >     // attributes, AND/OR attributes, and then calls SetMemorySpaceAttributes()
> >     // to set the new attributes.
> >     // So 0 CPU arch attributes should not happen as memory should always have
> >     // a cache attribute (no matter UC or WB, etc).
> >     //
> >     // Here, 0 CPU arch attributes will be filtered to be compatible with the
> >     // case that caller just calls SetMemorySpaceAttributes() with none CPU
> >     // arch attributes (for example, RUNTIME) as the purpose of the case is not
> >     // to clear CPU arch attributes.
> >     //
> >     if (CpuArchAttributes != 0) {
> >       if (gCpu == NULL) {
> >         Status = EFI_NOT_AVAILABLE_YET;
> >       } else {
> >         Status = gCpu->SetMemoryAttributes (
> >                          gCpu,
> >                          BaseAddress,
> >                          Length,
> >                          CpuArchAttributes
> >                          );
> >       }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The original call is doing exactly what should be done. However, in practice the attributes of the GCD memory space map are not currently accurate in Project Mu or EDK2, so passing in the attributes of the existing memory space descriptor won't work as intended.
> 
> Okay, I can see that. Perhaps in Project Mu we should update that comment then with some of these additional details. Do you think this is a broader statement that should be made there?

Do you mean update the following comment?
```
    // Caller is expected to use GetMemorySpaceDescriptor() to get the current
    // attributes, AND/OR attributes, and then calls SetMemorySpaceAttributes()
    // to set the new attributes.
```
If so, I do think updating it in Project Mu would be valuable. I can update the commits and make this a rebase PR.

I'll work on the necessary updates to keep the CPU arch attributes of the GCD memory map and EFI memory map in sync soon.

PR #59, opened 2022-11-15
Remove custom mode setting during PK deletion

kuqin12:
# Preface

Please ensure you have read the [contribution docs](https://github.com/microsoft/mu/blob/master/CONTRIBUTING.md) prior
to submitting the pull request. In particular,
[pull request guidelines](https://github.com/microsoft/mu/blob/master/CONTRIBUTING.md#pull-request-best-practices).

## Description

This change will remove the side effect of automatically setting the setup mode to `CustomMode` after deleting PK. The usage of `CustomMode` is largely avoided in the scope of Project Mu to reduce attack surfaces. This change is to align with the original behavior provided through `MsSecureBootLib`.

For each item, place an "x" in between `[` and `]` if true. Example: `[x]`.
_(you can also check items in the GitHub UI)_

- [x] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on Q35 virtual platform.

## Integration Instructions

N/A


SecurityPkg/SecurityPkg.dec
```diff
@@ -637,13 +637,21 @@
...
+# MU_CHANGE START
 [PcdsFeatureFlag]
+  # TCBZ2506
   ## Indicates if the platform will enforce PK's to be self signed when setting the PK with none currently set
   #   TRUE  - Require PK to be self signed
   #   FALSE - Does not require PK to be self signed
   # @Prompt Enforce PK to be self signed
   gEfiMdeModulePkgTokenSpaceGuid.PcdEnforceSelfsignedPk|FALSE|BOOLEAN|0x00010024
+
+  ## Indicates if the platform will set the custom secure boot mode variable during deletion of PK
```
makubacki: Does platform code here mean the code integrating SecureBootVariableLib in SecurityPkg?

Based on the code changes, I thought this PCD means "Indicates if the platform key deletion code in this package will set the custom secure boot mode variable...".

makubacki:
Given the PR is just changing the logic in one function. The function could technically take this as a parameter. Is that not done because the logic is needed elsewhere? And/or, not to change the function signature in the library interface?

kuqin12:
> Given the PR is just changing the logic in one function. The function could technically take this as a parameter. Is that not done because the logic is needed elsewhere? And/or, not to change the function signature in the library interface?

The real spirit is that we hope this will not be needed within the scope of MU. Maybe it is better to just comment out this section?

SecurityPkg/Library/SecureBootVariableLib/SecureBootVariableLib.c
```diff
@@ -578,9 +578,12 @@ DeletePlatformKey (
 {
   EFI_STATUS  Status;
 
-  Status = SetSecureBootMode (CUSTOM_SECURE_BOOT_MODE);
-  if (EFI_ERROR (Status)) {
-    return Status;
+  // MU_CHANGE: Remove the indicator of custom secure boot mode variable
+  if (FeaturePcdGet (PcdSupportCustomSecureBootModeVariable)) {
+    Status = SetSecureBootMode (CUSTOM_SECURE_BOOT_MODE);
```
Flickdm: So just to explain what I've learned about this,

CUSTOM_SECURE_BOOT_MODE / STANDARD_SECURE_BOOT_MODE (i.e "CustomMode") is not a concept defined in the UEFI Specification. STANDARD_SECURE_BOOT_MODE  is the implicit default mode as per the UEFI Specification. 

"CustomMode" is defined in the [EDKII implementation](https://cdrdv2-public.intel.com/671120/a-tour-beyond-bios-implementing-uefi-authenticated-variables-in-smm-with-edkii.pdf) and required by the [windows hardware certification requirements for client and server systems](https://learn.microsoft.com/en-us/previous-versions/windows/hardware/cert-program/windows-hardware-certification-requirements-for-client-and-server-systems?redirectedfrom=MSDN#:~:text=On%20an%20ARM%20system%2C%20it%20is%20forbidden%20to%20enable%20Custom%20Mode.%20Only%20Standard%20Mode%20may%20be%20enabled) and described by the [Signing UEFI Applications And Drivers for UEFI Secure Boot](https://github.com/tianocore-docs/Docs/raw/master/User_Docs/SigningUefiImages%20-v1dot31.pdf) documentation.

Quote (windows hardware certification requirements for client and server systems):
"On non-ARM systems, the platform MUST implement the ability for a physically present user to select between two Secure Boot modes in firmware setup: "Custom" and "Standard". Custom Mode allows for more flexibility as specified in the following: 
[...]

The firmware setup shall indicate if Secure Boot is turned on, and if it is operated in Standard or Custom Mode. The firmware setup must provide an option to return from Custom to Standard Mode which restores the factory defaults. On an ARM system, it is forbidden to enable Custom Mode. Only Standard Mode may be enabled."

So as long as this change, on any arm based system, leaves the device in STANDARD_SECURE_BOOT_MODE and any non-arm based system has this Pcd. Then this change does not conflict with any specification or requirement.

kuqin12: Thanks for the information. It has been a while, but from the conversation last time, the consensus was to not put in this section for good.

Given the change here is only "not setting to custom mode" before deleting the PK, I do not think this should not be considered a violation of the specs. But please correct me if I am wrong.

kuqin12:
> 

Okay, the PCD is not in place anymore. The custom mode is simply commented out.

SecurityPkg/SecurityPkg.dec
kuqin12: Thanks, this PCD should not be needed anymore.

apop5 (approved):

mikeytdisco (approved):

PR #60, opened 2022-11-16
Drop EmbeddedPkg from mu_tiano_plus

spbrogan:
For some hardware platform leveraging Project Mu there are conflicting changes in EmbeddedPkg. Given the limited investments by consuming silicon vendors, Project Mu and the Tianocore community in improving quality and getting alignment for the content from EmbeddedPkg, we will drop this package from mu_tiano_plus

- [X] Breaking change?
  - Will this change break pre-existing builds or functionality without action being taken?
 YES, for those platforms using EmbeddedPkg they will need to reference the new location in Project Mu.

## Integration Instructions

Add a new submodule to your project and update your PackagesPath


makubacki (approved):

makubacki:
I think it'd be helpful to update this with the actual location before submitting so it's visible in history:

"YES, for those platforms using EmbeddedPkg they will need to reference the new location in Project Mu."

spbrogan:
> I think it'd be helpful to update this with the actual location before submitting so it's visible in history:
> 
> "YES, for those platforms using EmbeddedPkg they will need to reference the new location in Project Mu."

i agree.  You have any good ideas.  
I really hate to setup a new repository, new builds, etc just for this package.

makubacki:
> > I think it'd be helpful to update this with the actual location before submitting so it's visible in history:
> > "YES, for those platforms using EmbeddedPkg they will need to reference the new location in Project Mu."
> 
> i agree. You have any good ideas. I really hate to setup a new repository, new builds, etc just for this package.

I agree the overhead is a lot for something we essentially don't want to keep in mu_tiano_plus.

Is there enough consumption from here that we need to keep the package as-is? Or could we break out a few important libraries/modules and move them somewhere else?

In that case, we would obviously lose some of the direct leverage of package-level rebases from EmbeddedPkg in edk2, but I think it's debatable if us focusing on improving the content we care about is better long term than trying to keep the whole package it in sync with edk2.

Just a suggestion to consider since I'm not sure what the consumption level looks like across vendor packages.

apop5 (approved):

makubacki:
@spbrogan, are you planning to drop this PR now?

makubacki:
Closing per offline discussion

PR #224, opened 2024-01-19
Partial Revert "SecurityPkg/SecureBoot: Support RSA4096 and RSA3072"

kenlautner:
## Description

Edk2 updated AuthVariable and secureboot to allow them to use SHA384 and SHA512.  The AuthVariable addition is good because it allows signing this with the PK but the secureboot addition is unnecessary.  

The secureboot change has things hashed by all three algorithms and then checking them in the DBX for SHA256, SHA384 and SHA512 lists to make sure it's not on any of them.  The issue with this is two fold.

1. This will have a performance impact.  One that many platforms will not want.
2. This is completely unnecessary because the only group putting things in the DBX is Microsoft and we only use SHA256.

For these reasons it makes sense to revert the change in the secureboot logic and keep the AuthVariable changes.
Commit in edk2 for reference: https://github.com/tianocore/edk2/commit/bbf182229587958b17336c114e0a1525c4f90f3d

- [] Impacts functionality?
  - **Functionality** - Does the change ultimately impact how firmware functions?
  - Examples: Add a new library, publish a new PPI, update an algorithm, ...
- [] Impacts security?
  - **Security** - Does the change have a direct security impact on an application,
    flow, or firmware?
  - Examples: Crypto algorithm change, buffer overflow fix, parameter
    validation improvement, ...
- [ ] Breaking change?
  - **Breaking change** - Will anyone consuming this change experience a break
    in build or boot behavior?
  - Examples: Add a new library class, move a module to a different repo, call
    a function in a new library class in a pre-existing module, ...
- [ ] Includes tests?
  - **Tests** - Does the change include any explicit test code?
  - Examples: Unit tests, integration tests, robot tests, ...
- [ ] Includes documentation?
  - **Documentation** - Does the change contain explicit documentation additions
    outside direct code modifications (and comments)?
  - Examples: Update readme file, add feature readme file, link to documentation
    on an a separate Web page, ...

## How This Was Tested

Tested on Intel Physical systems.  No issues seen.

## Integration Instructions

N/A

### DIFF
--- a/SecurityPkg/Library/DxeImageVerificationLib/DxeImageVerificationLib.c
+++ b/SecurityPkg/Library/DxeImageVerificationLib/DxeImageVerificationLib.c
@@ -1630,7 +1630,7 @@ Done:
       in the security database "db", and no valid signature nor any hash value of the image may
       be reflected in the security database "dbx".
     Otherwise, the image is not signed,
-      The hash value of the image must match a record in the security database "db", and
+      The SHA256 hash value of the image must match a record in the security database "db", and
       not be reflected in the security data base "dbx".
 
   Caution: This function may receive untrusted input.
@@ -1700,8 +1700,6 @@ DxeImageVerificationHandler (
   EFI_STATUS                    VarStatus;
   UINT32                        VarAttr;
   BOOLEAN                       IsFound;
-  UINT8                         HashAlg;
-  BOOLEAN                       IsFoundInDatabase;
 
   SignatureList     = NULL;
   SignatureListSize = 0;
@@ -1711,7 +1709,6 @@ DxeImageVerificationHandler (
   Action            = EFI_IMAGE_EXECUTION_AUTH_UNTESTED;
   IsVerified        = FALSE;
   IsFound           = FALSE;
-  IsFoundInDatabase = FALSE;
 
   //
   // Check the image type and get policy setting.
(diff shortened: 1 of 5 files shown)


makubacki:
@kenlautner, it would be helpful to elaborate a bit more here on what "this is unnecessary and can confuse other SecureBoot logic so we're reverting it" means in the PR description for others coming across the change.

.pytool/CISettings.py
makubacki, line 1: For clarity in source history, can you squash everything related to the revert in one commit and then have a single commit that updates the codeql stuff in `CISettings.py`?

kenlautner: I plan on removing the CISettings stuff.  I have a separate PR already up for that.  I'll clean this one up to just be the original revert.

makubacki: Thanks, that works too.

kuqin12 (approved):

makubacki (approved):

Flickdm (approved):

makubacki:
> @kenlautner, it would be helpful to elaborate a bit more here on what "this is unnecessary and can confuse other SecureBoot logic so we're reverting it" means in the PR description for others coming across the change.

Thanks. The context is much more helpful now.

kuqin12:
@kenlautner we still need this for 202311 release, right?

PR #254, opened 2024-02-29
Delete PK broken on latest integration due to not using CUSTOM MODE or UserPhysicallyPresent(..)

Flickdm:
# Preface

## Description

This reverts the revert where we reverted the update that supports SHA384 and SHA512. 

When ProcessVarWithPk(..) is called the expected code path is that you are in CUSTOM MODE and a UserPhysicallyPresent. Neither of which Project MU does or supports. So we end up falling down into VerifyTimeBasedPayloadAndUpdate(..). From there we fall into VerifyTimeBasedPayload and then finally we were depending on a special case where the size wasn't checked to hit the following line
 
 ```c
  // If the VariablePolicy engine is disabled, allow deletion of any authenticated variables.
  if ((PayloadSize == 0) && ((Attributes & EFI_VARIABLE_APPEND_WRITE) == 0) && !IsVariablePolicyEnabled ()) {
    VerifyStatus = TRUE;
    goto Exit;
  }
 ```
This would work in 202302 and before allow us to delete the PK. However in this [commit](https://github.com/microsoft/mu_tiano_plus/commit/bbf182229587958b17336c114e0a1525c4f90f3d) the logic to detect digest algorithm was changed and now prevents any payload with an invalid signature size (such as a PK Delete payload) from working.

History:

1. [Original Commit From EDK2](https://github.com/tianocore/edk2/commit/bbf182229587958b17336c114e0a1525c4f90f3d)
2. [Cherry-pick into MU_TIANO_PLUS](https://github.com/microsoft/mu_tiano_plus/commit/bbf182229587958b17336c114e0a1525c4f90f3d)
3. [Partial Revert to remove unnecessary logic](https://github.com/microsoft/mu_tiano_plus/commit/36b848b39cdfe644a22c50f5421f873f3015e81f)

Bug seen where PK cannot be deleted
See Issue #246 

1. [Revert "SecurityPkg/SecureBoot: Support RSA4096 and RSA3072" and follow up change](https://github.com/microsoft/mu_tiano_plus/pull/246/commits/04539e1012879eafd1c09230f3b1d8ad2de6c8e9)
2. [Revert "SecurityPkg/SecureBoot: Support RSA4096 and RSA3072"](https://github.com/microsoft/mu_tiano_plus/pull/246/commits/189cf02c951eafe1449091320289964cedcaaf7e)

This patch series contains the original two commits and the bug fix
1. [Cherry-pick into MU_TIANO_PLUS](https://github.com/microsoft/mu_tiano_plus/commit/bbf182229587958b17336c114e0a1525c4f90f3d)
2. [Partial Revert to remove unnecessary logic](https://github.com/microsoft/mu_tiano_plus/commit/36b848b39cdfe644a22c50f5421f873f3015e81f)
3. [SecurityPkg/SecureBoot: Support special case where PK is being deleted](https://github.com/microsoft/mu_tiano_plus/commit/49d7400b2697f778fae4eae6e3bd22c7ede3c4f1).

Effectively this adds a special case where if the signature is 0

For each item, place an "x" in between `[` and `]` if true. Example: `[x]`.
_(you can also check items in the GitHub UI)_

- [X] Impacts functionality?
  - **Functionality** - Allows for a PK to be deleted
- [ ] Impacts security?
  - **Security** - No
- [ ] Breaking change?
  - **Breaking change** - No
- [ ] Includes tests?
  - **Tests** - No
- [ ] Includes documentation?
  - **Documentation** - No

## How This Was Tested

1. Boot QemuQ35
2. Enable Secure Boot with the Microsoft Only Certificates
3. Boot to the Front Page
7. Go to the Security Tab - Reboot to Front Page
8. Disable Secure Boot by selecting None
9. If broken firmware will assert 
10. otherwise proceed

Additionally, confirmed that authenticated variables with valid signature data using the hash algorithms SHA256, SHA384 and SHA512 still work.

## Integration Instructions

N/A

### DIFF
--- a/SecurityPkg/Library/AuthVariableLib/AuthService.c
+++ b/SecurityPkg/Library/AuthVariableLib/AuthService.c
@@ -718,10 +718,13 @@ ProcessVarWithPk (
   // Init state of Del. State may change due to secure check
   //
   Del = FALSE;
-  if (  (InCustomMode () && UserPhysicalPresent ())
+  // MU_CHANGE [START] SecurityPkg/Secureboot: Support Fix delete PK failing bug
+  if (  ((InCustomMode () && UserPhysicalPresent ()) || (!IsVariablePolicyEnabled ()))
      || (  (mPlatformMode == SETUP_MODE)
         && !(FeaturePcdGet (PcdRequireSelfSignedPk) && IsPk)))
   {
+    // MU_CHANGE [END] SecurityPkg/Secureboot: Support Fix delete PK failing bug
+
     Payload     = (UINT8 *)Data + AUTHINFO2_SIZE (Data);
     PayloadSize = DataSize - AUTHINFO2_SIZE (Data);
     if (PayloadSize == 0) {
--- a/SecurityPkg/Library/AuthVariableLib/AuthService.c
+++ b/SecurityPkg/Library/AuthVariableLib/AuthService.c
@@ -29,12 +29,125 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #include <Protocol/VariablePolicy.h>
 #include <Library/VariablePolicyLib.h>
 
+#define SHA_DIGEST_SIZE_MAX  SHA512_DIGEST_SIZE
+
+/**
+  Retrieves the size, in bytes, of the context buffer required for hash operations.
+
+  If this interface is not supported, then return zero.
+
+  @return  The size, in bytes, of the context buffer required for hash operations.
+  @retval  0   This interface is not supported.
+
+**/
+typedef
+UINTN
+(EFIAPI *EFI_HASH_GET_CONTEXT_SIZE)(
+  VOID
+  );
+
+/**
+  Initializes user-supplied memory pointed by Sha1Context as hash context for
+  subsequent use.
+
+  If HashContext is NULL, then return FALSE.
+  If this interface is not supported, then return FALSE.
+
+  @param[out]  HashContext  Pointer to Hashcontext being initialized.
+
+  @retval TRUE   Hash context initialization succeeded.
+  @retval FALSE  Hash context initialization failed.
+  @retval FALSE  This interface is not supported.
+
+**/
+typedef
+BOOLEAN
+(EFIAPI *EFI_HASH_INIT)(
+  OUT  VOID  *HashContext
+  );
+
+/**
+  Digests the input data and updates Hash context.
+
+  This function performs Hash digest on a data buffer of the specified size.
+  It can be called multiple times to compute the digest of long or discontinuous data streams.
+  Hash context should be already correctly initialized by HashInit(), and should not be finalized
+  by HashFinal(). Behavior with invalid context is undefined.
+
+  If HashContext is NULL, then return FALSE.
+  If this interface is not supported, then return FALSE.
+
+  @param[in, out]  HashContext  Pointer to the Hash context.
+  @param[in]       Data         Pointer to the buffer containing the data to be hashed.
+  @param[in]       DataSize     Size of Data buffer in bytes.
+
+  @retval TRUE   SHA-1 data digest succeeded.
+  @retval FALSE  SHA-1 data digest failed.
+  @retval FALSE  This interface is not supported.
+
+**/
+typedef
+BOOLEAN
+(EFIAPI *EFI_HASH_UPDATE)(
+  IN OUT  VOID        *HashContext,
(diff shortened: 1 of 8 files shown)


makubacki (approved):

SecurityPkg/Library/AuthVariableLib/AuthService.c
```diff
@@ -605,10 +718,13 @@ ProcessVarWithPk (
   // Init state of Del. State may change due to secure check
   //
   Del = FALSE;
-  if (  (InCustomMode () && UserPhysicalPresent ())
+  // MU_CHANGE [START] SecurityPkg/Secureboot: Support Fix delete PK failing bug
+  if (  (!IsVariablePolicyEnabled ())
```
kuqin12: Would this change the behavior? I thought PP is still needed when the system is enlightened to do so?

SecurityPkg/Library/AuthVariableLib/AuthService.c
```diff
@@ -29,12 +29,125 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
...
+typedef struct {
+  UINT32                       HashSize;
+  EFI_HASH_GET_CONTEXT_SIZE    GetContextSize;
+  EFI_HASH_INIT                Init;
+  EFI_HASH_UPDATE              Update;
+  EFI_HASH_FINAL               Final;
+  VOID                         **HashShaCtx;
+  UINT8                        *OidValue;
+  UINTN                        OidLength;
+} EFI_HASH_INFO;
```
kuqin12, line 135: Is this really efi standard definition? Is so, should it live somewhere else? The same applies to the function pointer definitions.

SecurityPkg/Library/AuthVariableLib/AuthVariableLib.c
```diff
@@ -91,7 +93,7 @@ VARIABLE_ENTRY_PROPERTY  mAuthVarEntry[] = {
   },
 };
 
-VOID  **mAuthVarAddressPointer[9];
+VOID  **mAuthVarAddressPointer[11];
```
kuqin12, line 96: :)

kuqin12 (approved):

SecurityPkg/Library/AuthVariableLib/AuthService.c
Flickdm: So alternatively I could do the following:

```c
  if (  ((InCustomMode () && UserPhysicalPresent ()) || (!IsVariablePolicyEnabled ()))
   || (  (mPlatformMode == SETUP_MODE)
        && !(FeaturePcdGet (PcdRequireSelfSignedPk) && IsPk)))
  {
 ```
 
 Which would keep the existing behavior. The reason I removed InCustomMode(..) and UserPhysicallyPresent(..) is because we take a stance in Project MU that we won't enable CUSTOM MODE nor will we use UserPhysicallyPresent(..)
 
So InCustomMode(..) will always be FALSE, and even if the system is UserPhysicallyPresent(..) enlightend it won't matter because CUSTOM MODE will always be false (for Project MU).

Comment (makubacki) [reply SecurityPkg/Library/AuthVariableLib/AuthService.c]: I agree you could consider that condition as well to not deviate from the upstream (e.g. (!IsVariablePolicyEnabled() || (InCustomMode() && UserPhysicalPresent())) || ...).

Comment (makubacki) [reply SecurityPkg/Library/AuthVariableLib/AuthService.c]: I wrote that while you were also responding. I think this could go different ways depending on how the overall project wants to treat consideration of custom and physical presence. My understanding was that those APIs are in place similar to the upstream (not deprecated / removed in a way a platform wouldn't be able to use them) but specified negatively as a Mu platform decision.

If a wider stance is taken that will never be used, it might make sense to ASSERT if present, so platforms don't have to guess where they're being taken into account versus the upstream.

Comment (Flickdm) [reply SecurityPkg/Library/AuthVariableLib/AuthService.c]: Yeah I think your initial understanding is still true. I don't think we can affirmatively say that a downstream platform wont make that decision it's just that we don't recomend it.

There are other examples where similiar patterns are used such as ProcessVariable(..)

```c
  if ((!EFI_ERROR (Status)) && IsDeleteAuthVariable (OrgVariableInfo.Attributes, Data, DataSize, Attributes) && (UserPhysicalPresent () || !IsVariablePolicyEnabled ())) {

``` 

We might need a larger discussion / more changes if we want to take that wider stance

PR #284, opened 2024-06-25
[Rebase & FF] 202405: Renaming shell test app in FmpDevicePkg

VivianNK:
## Description
Renaming shell test app in FmpDevicePkg

---
Cherry picked the following commit:

974b5f148f

---

- [ ] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Changes from release/202311

## Integration Instructions

Upgrading from 202311 to 202405 -> No change for platform

Upgrading from edk2 -> 202405, change the path (if it exists) as follows:

`FmpDevicePkg/Test/UnitTest/Library/FmpDependencyLib/FmpDependencyLibUnitTestsUefi.inf`
  `FmpDevicePkg/Test/UnitTest/Library/FmpDependencyLib/FmpDependencyLibUnitTestApp.inf`

### DIFF
--- a/FmpDevicePkg/FmpDevicePkg.dsc
+++ b/FmpDevicePkg/FmpDevicePkg.dsc
@@ -174,7 +174,7 @@
   #
   # Add UEFI Target Based Unit Tests
   #
-  FmpDevicePkg/Test/UnitTest/Library/FmpDependencyLib/FmpDependencyLibUnitTestsUefi.inf
+  FmpDevicePkg/Test/UnitTest/Library/FmpDependencyLib/FmpDependencyLibUnitTestApp.inf
 
 [BuildOptions]
   *_*_*_CC_FLAGS = -D DISABLE_NEW_DEPRECATED_INTERFACES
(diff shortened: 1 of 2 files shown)


FmpDevicePkg/FmpDevicePkg.dsc
makubacki, line 1: I know this was the case in the original commit, but since this is Rb & FF, a commit message describing the change at a high-level should be present in the commit.

makubacki: It also looks like line endings may have been converted, that should be called out too.

makubacki (approved):

FmpDevicePkg/FmpDevicePkg.dsc
VivianNK: Converted to LF... was that a mistake from the original commit? Or why would the inf need LF line endings? Our other shell test apps in mu_plus have CRLF

makubacki: They should be CRLF. They may have been converted to CRLF later. For this commit, please keep them as CRLF.

os-d (approved):

PR #327, opened 2024-08-29
[CHERRY-PICK][REBASE & FF] Revert Mu Commits In Favor of edk2 Commits

os-d:
## Description

This PR is the current set of mu_tiano_plus commits I have upstreamed to edk2 from release/202405. Some of these had changes from edk2, so it is not a 1:1 revert to commit.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

N/A.

## Integration Instructions

N/A.

### DIFF
--- a/PrmPkg/PrmConfigDxe/PrmConfigDxe.c
+++ b/PrmPkg/PrmConfigDxe/PrmConfigDxe.c
@@ -152,10 +152,15 @@ SetRuntimeMemoryRangeAttributes (
       continue;
     }
 
+    // The memory space descriptor access attributes are not accurate. Don't pass
+    // in access attributes so SetMemorySpaceAttributes() doesn't update them.
+    // EFI_MEMORY_RUNTIME is not a CPU arch attribute, so calling
+    // SetMemorySpaceAttributes() with only it set will not clear existing page table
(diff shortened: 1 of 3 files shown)


apop5 (approved):

makubacki (approved):

PrmPkg/PrmConfigDxe/PrmConfigDxe.c
makubacki, line 1: I think it could go either way on whether the cherry-pick just fixes up the commit versus it following the revert. This is fine if preferred though.

os-d: I think the pattern of revert then cherry-pick is preferred because I think it will make future integrations easier. When we run cleansing fire, it will see the original Mu commit as empty and tell us to drop it. The cherry-pick from edk2 will either be empty, in which case git will drop, or if there was some massaging, we'll see it was a cherry-pick and act accordingly. I worry about if we just do fixup cherry-picks that cleansing fire will still report them as having active lines and so we may end up with extraneous commits in the next integration. This approach makes 2405 history less clean, but cleaner for future integrations.

PR #377, opened 2025-03-10
PRM support for non-existent MMIO ranges

RonnyHansen:
## Description

This is regarding PRM modules that are describing MMIO ranges.

When `PrmConfigDxe` is calling `GetMemorySpaceDescriptor()` with a memory range that is visible to the boot processor but has not been added to the memory map `GetMemorySpaceDescriptor()` will return `EFI_SUCCESS` and then return a memory descriptor indicating that the region is non-existent. This causes `SetRuntimeMemoryRangeAttributes()` to believe that the region has already been added to the memory map and will eventually cause an `ASSERT`.

This PR allows for `SetRuntimeMemoryRangeAttributes()` to treat a non-existent MMIO range the same as a range that triggered a `EFI_NOT_FOUND` error response from `GetMemorySpaceDescriptor()`.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Manual testing with non-existent MMIO range.

## Integration Instructions

N/A

### DIFF
--- a/PrmPkg/PrmConfigDxe/PrmConfigDxe.c
+++ b/PrmPkg/PrmConfigDxe/PrmConfigDxe.c
@@ -100,18 +100,22 @@ SetRuntimeMemoryRangeAttributes (
 
     Status2 = EFI_NOT_FOUND;
     Status  = gDS->GetMemorySpaceDescriptor (RuntimeMmioRanges->Range[Index].PhysicalBaseAddress, &Descriptor);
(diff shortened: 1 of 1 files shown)


os-d (approved):
I can take this to edk2 after merging here

makubacki (approved):

PR #393, opened 2025-04-15
Tcg2Dxe Pre-Install Event Log Lib

Raymond-MS:
## Description

Added new event log library for logging measurements before Tcg2Dxe has run and the protocol has been installed. This is there for platforms that need to log events that don't have a PEI phase for Tcg2PreUefiEventLogLib and to avoid reuse with this library which could break existing platforms. Library is a NULL instance and should do nothing unless overridden at the Platform level.

For details on how to complete these options and their meaning refer to [CONTRIBUTING.md](https://github.com/microsoft/mu/blob/HEAD/CONTRIBUTING.md).

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Verified by building SBSA and Q35 at the with updates to the .dsc to include the libraries. Note that SBSA does not support TPM and as such, did not include the Tcg2PreInstallEventLogLib. 

## Integration Instructions

Update the .dsc to include this new library

### DIFF
--- a/SecurityPkg/Include/Library/Tcg2PreInstallEventLogLib.h
+++ b/SecurityPkg/Include/Library/Tcg2PreInstallEventLogLib.h
@@ -1,0 +1,22 @@
+/** @file -- Tcg2PreInstallEventLogLib.h
+  This describes the interface that should be published by instances of the
+  Tcg2PreInstallEventLogLib. This library can be used to publish TPM EventLog
+  entries for measurements before the TCG2 protocol is installed.
+
+Copyright (c) Microsoft Corporation.
+SPDX-License-Identifier: BSD-2-Clause-Patent
+**/
+
+#ifndef TCG_2_PRE_INSTALL_EVENT_LOG_LIB_H_
+#define TCG_2_PRE_INSTALL_EVENT_LOG_LIB_H_
+
+/**
+  Create the EventLog entries.
+**/
+VOID
+EFIAPI
+CreateTcg2PreInstallEventLogEntries (
+  VOID
+  );
+
+#endif // TCG_2_DXE_EVENT_LOG_LIB_H_
--- a/SecurityPkg/Include/Library/Tcg2PreInstallEventLogLib.h
+++ b/SecurityPkg/Include/Library/Tcg2PreInstallEventLogLib.h
@@ -1,7 +1,8 @@
 /** @file -- Tcg2PreInstallEventLogLib.h
   This describes the interface that should be published by instances of the
   Tcg2PreInstallEventLogLib. This library can be used to publish TPM EventLog
-  entries for measurements before the TCG2 protocol is installed.
+  entries for measurements before the TCG2 protocol is installed and for
+  platforms that start at the DXE phase.
 
 Copyright (c) Microsoft Corporation.
 SPDX-License-Identifier: BSD-2-Clause-Patent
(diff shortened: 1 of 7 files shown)


SecurityPkg/Include/Library/Tcg2PreInstallEventLogLib.h
```diff
@@ -0,0 +1,22 @@
+/** @file -- Tcg2PreInstallEventLogLib.h
+  This describes the interface that should be published by instances of the
+  Tcg2PreInstallEventLogLib. This library can be used to publish TPM EventLog
+  entries for measurements before the TCG2 protocol is installed.
```
kuqin12: For platforms that start from DXE phase.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
```diff
@@ -54,6 +54,7 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #include <Library/DeviceStateLib.h>
 #include <Library/PanicLib.h>
 // MU_CHANGE [END]
+#include <Library/Tcg2PreInstallEventLogLib.h> // MU_CHANGE
```
kuqin12, line 57: nit: you can just move this line above by 1.

kuqin12 (approved):

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
Raymond-MS: Going to disagree here as the above MU_CHANGE is marked with a comment mentioning the reason it was added which is unrelated to this change, sorry. :(

kuqin12: Oh, good. Here is fine then.

kuqin12 (approved):

eeshanl (approved):
LGTM, assume this has also been tested on a physical platform as well as qemu?

Raymond-MS:
> LGTM, assume this has also been tested on a physical platform as well as qemu?

Yes, I pulled the branch down at our platform level and made the necessary changes to SBSA and Q35 .dsc files and it builds successfully on  both.

eeshanl (approved):

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
```diff
@@ -3036,6 +3037,9 @@ DriverEntry (
     EfiCreateProtocolNotifyEvent (&gEfiResetNotificationProtocolGuid, TPL_CALLBACK, OnResetNotificationInstall, NULL, &Registration);
   }
 
+  // MU_CHANGE - Add support for measurements before the Tcg2 protocol is installed
+  CreateTcg2PreInstallEventLogEntries ();
```
apop5, line 3041: I have concerns here.

This is a null function that is called prior to the tcg2 protocol being installed. What is the expectation for this function? How will it perform the actions necessary? 

If the library implementation expects to reach into the internals of the tcg2 dxe driver then this will be coupling itself to the implementation of the dxe driver. This will cause the library to be brittle as the implementation of the tcg2 dxe can change independently. 

Is there another route that can be taken where the things that need added to the tcg2 log can be provided and consumed by the tcg2dxe ie just the data needs added to the log instead of injecting code that would add the data?

Raymond-MS: The concern is valid. I believe I found an alternative solution which will make everyone happy. :)

PR #395, opened 2025-04-17
[CHERRY PICK] Security Patch for CVE-2024-38797

Flickdm:
# Description

CVE: CVE-2024-38797

In HashPeImageByType(), the hash of a PE/COFF image is calculated. This function may get untrusted input.

Inside this function, the following code verifies the loaded image has the correct format, by reading the second byte of the buffer.

```c
  if ((*(AuthData + 1) & TWO_BYTE_ENCODE) != TWO_BYTE_ENCODE) {
  	...
  }
```

Since the input image is not trusted, it may not have a second byte present to read. So this poses a potential out of bounds read error.

With the below fix, we are ensuring that we don't perform an out of bound read. i.e, we make sure that AuthDataSize is greater than 1.

```c
  if (AuthDataSize > 1 && (*(AuthData + 1) & TWO_BYTE_ENCODE) != TWO_BYTE_ENCODE){
    ...
  }
```

AuthDataSize size is verified before reading the second byte. So, if AuthDataSize is less than 2, the second byte will not be read, and the out of bound read will be prevented.

---

The logic of the function was cleaned up to provide an early exit if the untrusted data does not have enough bytes to check for a TWO_BYTE_ENCODE or if the TWO_BYTE_ENCODE is not present. 

This is slightly more efficient than checking this condition in the for loop.

Additionally, when the hash algorithm selected by Index has such a large OID that the OID Comparison couldn't be performed - this would lead to an early break rather than a OID Mismatch. This has been changed to a `continue` instead.

Original PR https://github.com/tianocore/edk2/pull/10928
---

- [ ] Breaking change?
- [X] Impacts security?
   - Corrects out of bound read in HashPeImageByType ()
- [ ] Includes tests?

## How This Was Tested

Tested the patch on real platform with and without TPM connected and verified image is booting fine.

## Integration Instructions

N/A

### DIFF
--- a/SecurityPkg/VariableAuthenticated/SecureBootConfigDxe/SecureBootConfigImpl.c
+++ b/SecurityPkg/VariableAuthenticated/SecureBootConfigDxe/SecureBootConfigImpl.c
@@ -2114,30 +2114,35 @@ HashPeImageByType (
 {
   UINT8                     Index;
   WIN_CERTIFICATE_EFI_PKCS  *PkcsCertData;
+  UINT32                    PkcsCertSize;
 
   PkcsCertData = (WIN_CERTIFICATE_EFI_PKCS *)(mImageBase + mSecDataDir->Offset);
+  PkcsCertSize = mSecDataDir->SizeOfCert;
+
+  //
+  // Check the Hash algorithm in PE/COFF Authenticode.
+  //    According to PKCS#7 Definition:
+  //        SignedData ::= SEQUENCE {
+  //            version Version,
+  //            digestAlgorithms DigestAlgorithmIdentifiers,
+  //            contentInfo ContentInfo,
+  //            .... }
+  //    The DigestAlgorithmIdentifiers can be used to determine the hash algorithm in PE/COFF hashing
+  //    This field has the fixed offset (+32) in final Authenticode ASN.1 data.
+  //    Fixed offset (+32) is calculated based on two bytes of length encoding.
+  //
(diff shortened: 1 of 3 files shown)


SecurityPkg/VariableAuthenticated/SecureBootConfigDxe/SecureBootConfigImpl.c
```diff
@@ -2114,30 +2114,35 @@ HashPeImageByType (
...
-    //    The DigestAlgorithmIdentifiers can be used to determine the hash algorithm in PE/COFF hashing
-    //    This field has the fixed offset (+32) in final Authenticode ASN.1 data.
-    //    Fixed offset (+32) is calculated based on two bytes of length encoding.
+    // Only support two bytes of Long Form of Length Encoding.
     //
-    if ((*(PkcsCertData->CertData + 1) & TWO_BYTE_ENCODE) != TWO_BYTE_ENCODE) {
-      //
-      // Only support two bytes of Long Form of Length Encoding.
-      //
+    return EFI_BAD_BUFFER_SIZE;
```
kenlautner, line 2138: EFI_BAD_BUFFER_SIZE should be added to the retvals.  I don't know if we care enough to carry the difference from edk2 though.

kenlautner (approved):

apop5 (approved):

makubacki (approved):

PR #396, opened 2025-04-17
Update MuTcg2Protocol

Raymond-MS:
## Description

Updated the MuTcg2Protocol to include a function to initialize an event. This is used alongside Tcg2LogEvent to log and NOT extend events.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Built and tested using a driver to register for the protocol to be installed. Once installed, a callback is invoked which logs events. 

## Integration Instructions

N/A

### DIFF
--- a/SecurityPkg/Include/Protocol/MuTcg2Protocol.h
+++ b/SecurityPkg/Include/Protocol/MuTcg2Protocol.h
@@ -40,8 +40,8 @@ EFI_STATUS
   );
 
 /**
-  The EFI_MU_TCG2_PROTOCOL MuEventInit function call provides callers with
-  an interface for initializing events before they are logged.
+  Initializes the provided TCG PCR Event2 Header structure based on its event type
+  so it is populated for further use such as logging.
 
   Supported Event Types:
   EV_NO_ACTION
--- a/SecurityPkg/Include/Protocol/MuTcg2Protocol.h
+++ b/SecurityPkg/Include/Protocol/MuTcg2Protocol.h
@@ -58,6 +58,7 @@ EFI_STATUS
 typedef
 EFI_STATUS
 (EFIAPI *MU_TCG2_INIT_EVENT)(
+  IN MU_TCG2_PROTOCOL        *This,
   IN OUT TCG_PCR_EVENT2_HDR  *Event,
   IN UINT32                  EventSize
   );
--- a/SecurityPkg/Include/Protocol/MuTcg2Protocol.h
+++ b/SecurityPkg/Include/Protocol/MuTcg2Protocol.h
@@ -12,7 +12,7 @@
 #include <IndustryStandard/Tpm20.h>
 #include <Protocol/Tcg2Protocol.h>
 
-#define MU_TCG2_PROTOCOL_VERSION  1
+#define MU_TCG2_PROTOCOL_VERSION  2
 
 typedef struct tdMU_TCG2_PROTOCOL MU_TCG2_PROTOCOL;
 
@@ -39,6 +39,22 @@ EFI_STATUS
   IN EFI_TCG2_EVENT       *Event
   );
 
+/**
+  The EFI_MU_TCG2_PROTOCOL MuEventInit function call provides callers with
+  an interface for initializing events before they are logged.
+
+  Supported Event Types:
+  EV_NO_ACTION
+  EV_EFI_HCRTM_EVENT
+
+  @param[in]      This             Indicates the calling context
+  @param[in/out]  Event            Pointer to the event to initialize
+  @param[in]      EventSize        Size of the event
+
+  @retval EFI_SUCCESS            Operation completed successfully.
+  @retval EFI_INVALID_PARAMETER  One or more of the parameters are incorrect.
+  @retval EFI_UNSUPPORTED        The event provided is unsupported
+**/
 typedef
 EFI_STATUS
 (EFIAPI *MU_TCG2_INIT_EVENT)(
--- a/SecurityPkg/Include/Protocol/MuTcg2Protocol.h
+++ b/SecurityPkg/Include/Protocol/MuTcg2Protocol.h
@@ -38,11 +38,19 @@ EFI_STATUS
   IN TPML_DIGEST_VALUES   *DigestList,
   IN EFI_TCG2_EVENT       *Event
   );
+
+typedef
+EFI_STATUS
+(EFIAPI *MU_TCG2_INIT_EVENT)(
+  IN OUT TCG_PCR_EVENT2_HDR  *Event,
+  IN UINT32                  EventSize
+  );
 // MU_CHANGE - END - Add a new protocol to support Log-only events.
 
 struct tdMU_TCG2_PROTOCOL {
-  UINT32               Version;
-  MU_TCG2_LOG_EVENT    Tcg2LogEvent;
+  UINT32                Version;
+  MU_TCG2_LOG_EVENT     Tcg2LogEvent;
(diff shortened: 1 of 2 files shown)


SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
```diff
@@ -1446,6 +1446,37 @@ Tcg2LogEvent (
...
+    goto Exit;
+  }
+
+  Status = EFI_SUCCESS;
+  if ((Event->EventType == EV_NO_ACTION) || (Event->EventType == EV_EFI_HCRTM_EVENT)) {
+    EventType = Event->EventType;
+    InitNoActionEvent (Event, EventSize);
+    Event->EventType = EventType;
+  } else {
+    // TODO: Add init for other event types
```
kuqin12, line 1489: Return unsupported?

kuqin12 (approved):

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
Raymond-MS: That is fine with me. :)

SecurityPkg/Include/Protocol/MuTcg2Protocol.h
```diff
@@ -38,11 +38,19 @@ EFI_STATUS
...
+(EFIAPI *MU_TCG2_INIT_EVENT)(
+  IN OUT TCG_PCR_EVENT2_HDR  *Event,
+  IN UINT32                  EventSize
+  );
 // MU_CHANGE - END - Add a new protocol to support Log-only events.
 
 struct tdMU_TCG2_PROTOCOL {
-  UINT32               Version;
-  MU_TCG2_LOG_EVENT    Tcg2LogEvent;
+  UINT32                Version;
```
makubacki, line 68: Is there a reason not to update the protocol version? Consumers need something (either a version update or a protocol GUID update) to know if the API is present.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
```diff
@@ -1446,6 +1446,37 @@ Tcg2LogEvent (
...
+EFIAPI
+Tcg2InitEvent (
+  IN OUT TCG_PCR_EVENT2_HDR  *Event,
+  IN UINT32                  EventSize
+  )
+{
+  EFI_STATUS     Status;
+  TCG_EVENTTYPE  EventType;
+
+  DEBUG ((DEBUG_VERBOSE, "%a - Entry\n", __FUNCTION__));
```
makubacki:
```suggestion
  DEBUG ((DEBUG_VERBOSE, "%a - Entry\n", __func__));
```

SecurityPkg/Include/Protocol/MuTcg2Protocol.h
```diff
@@ -38,11 +38,19 @@ EFI_STATUS
   IN TPML_DIGEST_VALUES   *DigestList,
   IN EFI_TCG2_EVENT       *Event
   );
+
+typedef
```
makubacki, line 58: Need a function description comment.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
```diff
@@ -1446,6 +1446,37 @@ Tcg2LogEvent (
...
+
+  if (Event == NULL) {
+    Status = EFI_INVALID_PARAMETER;
+    goto Exit;
+  }
+
+  Status = EFI_SUCCESS;
+  if ((Event->EventType == EV_NO_ACTION) || (Event->EventType == EV_EFI_HCRTM_EVENT)) {
```
makubacki, line 1484: Supported event types need to be described in the API description comment.

SecurityPkg/Include/Protocol/MuTcg2Protocol.h
makubacki: Had to leave the comment here due to modified lines, but I'm referring to updating `MU_TCG2_PROTOCOL_VERSION` above in the file.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
Raymond-MS: Would you recommend I make the change to the above Tcg2LogEvent function as well? I was using it as an example. :)

makubacki: Yes, please. Anything in the surrounding area in `MU_CHANGE` tags is fine to get on the new format.

SecurityPkg/Include/Protocol/MuTcg2Protocol.h
```diff
@@ -38,11 +38,35 @@ EFI_STATUS
   IN TPML_DIGEST_VALUES   *DigestList,
   IN EFI_TCG2_EVENT       *Event
   );
+
+/**
+  The EFI_MU_TCG2_PROTOCOL MuEventInit function call provides callers with
+  an interface for initializing events before they are logged.
```
makubacki: I suggest you drop the symbol name details. The description itself will usually be associated with the code (e.g. in Doxygen and VS Code) and it's easier to read just the description.

```suggestion
  Initializes a TCG PCR Event structure so it is populated for further use such as logging.
```

SecurityPkg/Include/Protocol/MuTcg2Protocol.h
```diff
@@ -38,11 +38,35 @@ EFI_STATUS
...
+
+/**
+  The EFI_MU_TCG2_PROTOCOL MuEventInit function call provides callers with
+  an interface for initializing events before they are logged.
+
+  Supported Event Types:
+  EV_NO_ACTION
+  EV_EFI_HCRTM_EVENT
+
+  @param[in]      This             Indicates the calling context
```
makubacki, line 50: There is not a `This` parameter on the actual function signature. There should be for consistency across the protocol API.

SecurityPkg/Include/Protocol/MuTcg2Protocol.h
makubacki: I know this is based on the example above. You don't need to modify it.

makubacki: I also updated my suggestion to include a clarification that this only modifies the `Event` structure provided. I think that's helpful because the actual operation the function performs feels like something that would be in more of a helper library versus a protocol.

makubacki (approved):

makubacki:
We decided to pursue a more standardized and sustainable option for the changes that will lead to @Raymond-MS abandoning this PR.

1. Move the `MU_TCG2_LOG_EVENT` outside of `MU_TCG2_PROTOCOL`
   - When that is done deprecate `MU_TCG2_PROTOCOL` in favor of the spec-defined version
   - Current suggestion is to propose adding the interface to `EFI_TCG2_PROTOCOL` via USWG ECR and the move the code to edk2 under the code first process
2. Move the function being added in this PR (`MU_TCG2_INIT_EVENT`) to a library. It only depends on `InitNoActionEvent()` in Tcg2Dxe.c, but that's a relatively simple function with no other dependencies in the module. Consider moving it to the library as well and eliminate dynamic dispatch complexity from being necessary for that API. This can proceed without a spec change to the upstream.

PR #399, opened 2025-04-22
Tcg2Dxe Protocol Updates + Tcg2InitEventLib Creation

Raymond-MS:
## Description

Added Tcg2LogEvent to the Tcg2Protocol which is used to logging events without extending to the TPM. Created a library (Tcg2InitEventLib) which is used for initializing events for logging purposes. Removed MuTcg2Protocol which was used to only log events.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Successfully built and tested the protocol using a driver to log events

## Integration Instructions

N/A


SecurityPkg/Include/Library/Tcg2InitEventLib.h
```diff
@@ -0,0 +1,57 @@
...
+**/
+
+#ifndef TCG_2_INIT_EVENT_LIB_H_
+#define TCG_2_INIT_EVENT_LIB_H_
+
+#include <IndustryStandard/UefiTcgPlatform.h>
+
+/**
+  This function initialize TCG_PCR_EVENT2_HDR for EV_NO_ACTION Event Type other than EFI Specification ID event
+  The behavior is defined by TCG PC Client PFP Spec. Section 9.3.4 EV_NO_ACTION Event Types
```
makubacki, line 16: This is a little much to take in all at once. Does this still capture what is being stated?

```suggestion
  This function initializes the TCG_PCR_EVENT2_HDR for EV_NO_ACTION events.
  
  Note: The EFI Specification ID event is not supported.

  The behavior is defined by TCG PC Client PFP Spec. Section 9.3.4 EV_NO_ACTION Event Types.
```

SecurityPkg/Include/Library/Tcg2InitEventLib.h
```diff
@@ -0,0 +1,57 @@
...
+
+  It is considered an error to call this function with a NULL pointer, or an ActivePcrBanks value of 0.
+
+  @param[in, out]   Event           Event Header of EV_NO_ACTION Event
```
makubacki, line 20: To reduce potential confusion that this is a C type contradictory with the actual parameter type (`TCG_PCR_EVENT2_HDR`).

```suggestion
  @param[in, out]   Event           Event Header of an EV_NO_ACTION type event
```

SecurityPkg/Include/Library/Tcg2InitEventLib.h
```diff
@@ -0,0 +1,57 @@
...
+  @param[in]        EventSize       Event Size of the EV_NO_ACTION Event
```
makubacki, line 21:
```suggestion
  @param[in]        EventSize       Size of the Event argument

```

SecurityPkg/Library/Tcg2InitEventLib/Tcg2InitEventLib.c
```diff
@@ -0,0 +1,148 @@
...
+  )
+{
+  UINT32         DigestListCount;
+  TPMI_ALG_HASH  HashAlgId;
+  UINT8          *DigestBuffer;
+
+  DEBUG ((DEBUG_VERBOSE, "InitNoActionEvent ...\n"));
+
+  if (Event == NULL) {
+    DEBUG ((DEBUG_ERROR, "Event parameter is NULL\n"));
```
makubacki, line 49: To reduce code verbosity, I'd let the caller decide what to do with the error codes including printing. If they are kept, they should include the function name so they can be traced more easily to code from the serial log since they're generic messages.

SecurityPkg/Library/Tcg2InitEventLib/Tcg2InitEventLib.c
```diff
@@ -0,0 +1,148 @@
...
+  // Set Digests Count
+  //
+  WriteUnaligned32 ((UINT32 *)&Event->Digests.count, DigestListCount);
+
+  //
+  // Set Event Size
+  //
+  WriteUnaligned32 ((UINT32 *)DigestBuffer, EventSize);
+
+  DEBUG ((DEBUG_VERBOSE, "InitNoActionEvent - %r\n", EFI_SUCCESS));
```
makubacki, line 114: Even in verbose, I don't think this print is necessary. But feel to keep if you like it.

SecurityPkg/Library/Tcg2InitEventLib/Tcg2InitEventLib.c
```diff
@@ -0,0 +1,148 @@
...
+
+  return EFI_SUCCESS;
+}
+
+/**
+  This function initialize TCG_PCR_EVENT2_HDR for HCRTM Event Type.
```
makubacki, line 120:
```suggestion
  This function initializes a TCG_PCR_EVENT2_HDR for a HCRTM type event.
```

SecurityPkg/Library/Tcg2InitEventLib/Tcg2InitEventLib.inf
```diff
@@ -0,0 +1,37 @@
...
+  MdePkg/MdePkg.dec
+  MdeModulePkg/MdeModulePkg.dec
+  SecurityPkg/SecurityPkg.dec
+
+[LibraryClasses]
+  DebugLib
+  BaseLib
+  UefiLib
+  UefiBootServicesTableLib
+  Tpm2CommandLib
```
makubacki, line 37: Please sort since this is a new library instance.

SecurityPkg/Library/Tcg2InitEventLib/Tcg2InitEventLib.inf
```diff
@@ -0,0 +1,37 @@
...
+  BASE_NAME           = Tcg2InitEventLib
+  FILE_GUID           = E64B8E39-818D-4681-8797-55F96432FDA6
+  VERSION_STRING      = 1.0
+  MODULE_TYPE         = BASE
+  LIBRARY_CLASS       = Tcg2InitEventLib
+
+#
+# The following information is for reference only and not required by the build tools.
+#
+#  VALID_ARCHITECTURES           = IA32 X64 AARCH64
```
makubacki, line 21: Since this is targeting upstream, I'd drop the `VALID_ARCHITECTURES` section. Nothing here restricts the code to these architectures, and it won't stay accurate into the future.

SecurityPkg/Library/Tcg2InitEventLib/Tcg2InitEventLib.c
```diff
@@ -0,0 +1,148 @@
+/** @file
+  This instance of the Tcg2InitEventLib is available to initialize events
+  for logging purposes.
```
makubacki, line 3: There is no need to couple an assumption about how the library will be used.

```suggestion
  Initializes TCG event data structures.
```

SecurityPkg/SecurityPkg.dec
```diff
@@ -136,6 +136,12 @@
   #   a user with Physical Presence requests
   #
   Tcg2PhysicalPresencePromptLib|Include/Library/Tcg2PhysicalPresencePromptLib.h
+
+  ##  @libraryclass  This library provides an interface to initialize TCG2 events
+  #   for logging purposes
```
makubacki, line 141:
```suggestion
  ##  @libraryclass  Initializes TCG event data structures.
  #
```

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
```diff
@@ -1410,15 +1332,15 @@ Tcg2HashLogExtendEvent (
...
+  IN EFI_TCG2_PROTOCOL   *This,
   IN TPML_DIGEST_VALUES  *DigestList,
   IN EFI_TCG2_EVENT      *Event
   )
 {
   EFI_STATUS         Status;
   TCG_PCR_EVENT_HDR  NewEventHdr;
 
-  DEBUG ((DEBUG_VERBOSE, "%a - Entry\n", __FUNCTION__));
+  DEBUG ((DEBUG_VERBOSE, "Tcg2LogEvent ...\n"));
```
makubacki, line 1343: Was there a reason to change this from the old message? To make this PR match closely to what will go upstream, please move any "Mu refactoring" unrelated to Tcg2InitEventLib to a separate commit.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
```diff
@@ -1614,17 +1534,11 @@ Tcg2GetResultOfSetActivePcrBanks (
...
 }
 
-// MU_CHANGE - START - Add a new protocol to support Log-only events.
-MU_TCG2_PROTOCOL  mMuTcg2Protocol = {
-  MU_TCG2_PROTOCOL_VERSION,
-  Tcg2LogEvent
-};
-// MU_CHANGE - END - Add a new protocol to support Log-only events.
-
 EFI_TCG2_PROTOCOL  mTcg2Protocol = {
```
makubacki, line 1537: This function cannot go into the middle of the `EFI_TCG2_PROTOCOL` structure. The protocol interface is a binary interface, and this is not backward compatible. It must go at the end of the structure.

makubacki, line 1537: What will the guidance be for `EFI_TCG2_PROTOCOL` consumers to know if this function is present? Because protocols are a binary interface, driver writers will need to be able to determine if the protocol producer in their platform has the function or not. Usually, a protocol version field would be used for this. I don't see one on `EFI_TCG2_PROTOCOL`.

As disruptive as it may be, if there is no other solution, it a V2 `EFI_TCG2_PROTOCOL` could be proposed with a new GUID, a version field in the protocol interface structure, and this function present.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
makubacki, line 1: I know these changes are going to quickly target upstream, but until then, they should still technically have `MU_CHANGE` tags. Changes might not end up matching the upstream 1:1 and that will make it a bit more obvious in the source code which version is present.

SecurityPkg/Library/Tcg2InitEventLib/Tcg2InitEventLib.c
Raymond-MS: This was mainly for debugging purposes. Doesn't really need to be here, will remove.

SecurityPkg/Library/Tcg2InitEventLib/Tcg2InitEventLib.c
Raymond-MS: This was mainly for debugging purposes, will remove.

SecurityPkg/Library/Tcg2InitEventLib/Tcg2InitEventLib.c
Raymond-MS: Going to keep this in the same format as the InitNoActionEvent:

This function initializes the TCG_PCR_EVENT2_HDR for HCRTM events.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
Raymond-MS: Yes, because this was under MuTcg2Protocol, this was in an entirely different style than any other protocol in this file. I made the change to more adhere to the standard set by the other protocol functions.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
Raymond-MS: Let's go with a V2 of the protocol. I think that is the safest and most straightforward path to take at the moment. This way we can leave the current protocol alone and let those who want to move to V2 do so.

PR #401, opened 2025-04-25
[REBASE + FF] Add TpmLogProtocol + Tcg2InitEventLib + Remove MuTcg2Protocol

Raymond-MS:
## Description

Added Tcg2InitEventLib which is a library used to initialize events for logging. Added a new TpmLogProtocol which is meant to log events without extending to the TPM. Removed MuTcg2Protocol.

For details on how to complete these options and their meaning refer to [CONTRIBUTING.md](https://github.com/microsoft/mu/blob/HEAD/CONTRIBUTING.md).

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Verified functionality by testing calls to the TpmLogProtocol from a DXE Driver.

## Integration Instructions

Removal of MuTcg2Protocol can be a breaking change. If this change causes a break in your build, please send an e-mail to myself or any other MU maintainer such that we can work with you to get it resolved. 

If possible, you should be able to replace MuTcg2Protocol with the new TpmLogProtocol which has similar functionality just with a different GUID.


os-d:
Can you make this a [REBASE & FF] PR and have separate commits to check in, one of them being a direct revert of the MuTcg2Protocol commit (and any other commits updating it reverted).

Also, can you add the integration instructions since this is a breaking change? Partners consume this from the release notes that get autogenerated.

SecurityPkg/Library/Tcg2InitEventLib/Tcg2InitEventLib.c
```diff
@@ -0,0 +1,143 @@
...
+  Event->EventType = EV_NO_ACTION;
+
+  //
+  // Set Hash count & hashAlg accordingly, while Digest.digests[n].digest to all 0
+  //
+  ZeroMem (&Event->Digests, sizeof (Event->Digests));
+
+  if ((ActivePcrBanks & EFI_TCG2_BOOT_HASH_ALG_SHA1) != 0) {
+    HashAlgId = TPM_ALG_SHA1;
+    CopyMem (DigestBuffer, &HashAlgId, sizeof (TPMI_ALG_HASH));
```
os-d: nit: all the sizeofs in this section to be `sizeof (HashAlgId)` so if that changes type in the future it isn't broken, also makes it a little clearer why that size is being used.

SecurityPkg/Library/Tcg2InitEventLib/Tcg2InitEventLib.c
```diff
@@ -0,0 +1,143 @@
+/** @file
```
os-d, line 1: Is this library intended to be used outside of Tcg2Dxe? If not, should it just be part of Tcg2Dxe? There are already a lot of Tcg2 libs :).

os-d (approved):

SecurityPkg/Library/Tcg2InitEventLib/Tcg2InitEventLib.c
Raymond-MS: Yes, the entire reason for the creation of the library was so that it could be used outside of Tcg2Dxe. There are other drivers that may need to initialize an event/digest for use with the new log protocol.

os-d: Gotcha, I was thinking this was being used on the protocol side, but I see it is a helper for consumers of the protocol.

SecurityPkg/Include/Library/Tcg2InitEventLib.h
```diff
@@ -0,0 +1,59 @@
+/** @file
+  Initializes TCG event data structures
+
+Copyright (c) Microsoft Corporation.
+SPDX-License-Identifier: BSD-2-Clause-Patent
+**/
```
makubacki, line 6: You can consider adding `MU_CHANGE` tags to the top of the file. It sometimes helps connect the pieces.

SecurityPkg/Library/Tcg2InitEventLib/Tcg2InitEventLib.c
```diff
@@ -0,0 +1,143 @@
...
+
+**/
+
+#include <Uefi.h>
+#include <Library/Tpm2CommandLib.h>
+#include <Library/DebugLib.h>
+#include <Library/BaseLib.h>
+#include <Library/UefiLib.h>
+#include <Library/UefiLib/UefiLibInternal.h>
+#include <Library/UefiBootServicesTableLib.h>
```
makubacki, line 15: Is `UefiBootServicesTableLib` actually needed?

makubacki: Please double check others too. A `BASE` library shouldn't have a dependency on something like boot services.

makubacki (approved):

SecurityPkg/Include/Library/Tcg2InitEventLib.h
```diff
@@ -0,0 +1,59 @@
...
+  @retval EFI_SUCCESS            Operation completed successfully.
+  @retval EFI_INVALID_PARAMETER  One or more of the parameters are incorrect.
+
+**/
+EFI_STATUS
+EFIAPI
+InitHcrtmEvent (
+  IN OUT TCG_PCR_EVENT2_HDR           *Event,
+  IN UINT32                           EventSize,
+  IN EFI_TCG2_EVENT_ALGORITHM_BITMAP  ActivePcrBanks
```
spbrogan, line 56: why do you need active pcr banks.  
it is generally suggested to log events to all banks but if this is a non measured event that what does it matter what bank it goes into?

SecurityPkg/Include/Library/Tcg2InitEventLib.h
makubacki, line 1: The uncrustify commit should not be a part of Rebase + FF:

image

The "Protocol Name Change" commit should include the name of the original protocol in the commit subject to make it easy to locate in history.

SecurityPkg/Include/Library/Tcg2InitEventLib.h
```diff
@@ -0,0 +1,59 @@
+/** @file
+  Initializes TCG event data structures
```
spbrogan, line 2: I don't understand this library, who would use it and why?  Is it just a helper to init the caller supplied memory with HDR?   Is that hard to do and worth a library? 
Is HCRTM defined in a spec and is the plan to implement this to spec?

SecurityPkg/Include/Library/Tcg2InitEventLib.h
Raymond-MS: Yes, of course. Oliver helped me to adjust the commits. This won't be the case. No worries. :)

SecurityPkg/Library/Tcg2InitEventLib/Tcg2InitEventLib.c
```diff
@@ -0,0 +1,143 @@
...
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#include <Uefi.h>
+#include <Library/Tpm2CommandLib.h>
+#include <Library/DebugLib.h>
+#include <Library/BaseLib.h>
+#include <Library/UefiLib.h>
+#include <Library/UefiLib/UefiLibInternal.h>
```
spbrogan, line 14: Library/UefiLib/UefiLibInternal.h should not be used here.  It is not a public header file

SecurityPkg/Library/Tcg2InitEventLib/Tcg2InitEventLib.c
```diff
@@ -0,0 +1,143 @@
...
+/**
+  This function initializes the TCG_PCR_EVENT2_HDR for EV_NO_ACTION events. The behavior is defined
+  by TCG PC Client PC Client PFP SPec. Secion 9.3.4 EV_NO_ACTION Event Types.
+
+  Note: The EFI Specification ID event is not supported.
+
+  It is considered an error to call this function with a NULL pointer, or an ActivePcrBanks value of 0.
+
+  @param[in, out]   Event           Event Header of an EV_NO_ACTION type event
+  @param[in]        EventSize       Size of the Event argument
```
spbrogan, line 28: this implementation never checks EventSize to make sure Event is large enough of a memory allocation.

SecurityPkg/Library/Tcg2InitEventLib/Tcg2InitEventLib.c
```diff
@@ -0,0 +1,143 @@
...
+  //
+  // Set Event Size
+  //
+  WriteUnaligned32 ((UINT32 *)DigestBuffer, EventSize);
+
+  return EFI_SUCCESS;
+}
+
+/**
+  This function initializes the TCG_PCR_EVENT2_HDR for HCRTM events.
```
spbrogan, line 115: This function doesn't make sense to me.  will talk offline

SecurityPkg/Library/Tcg2InitEventLib/Tcg2InitEventLib.inf
```diff
@@ -0,0 +1,30 @@
+## @file Tcg2InitEventLib.inf
+# Initializes TCG event data structures
```
spbrogan, line 2: need better desciprtion of the library

SecurityPkg/SecurityPkg.dsc
```diff
@@ -271,6 +272,7 @@
   SecurityPkg/Library/Tcg2PpVendorLibNull/Tcg2PpVendorLibNull.inf
   SecurityPkg/Library/TcgPpVendorLibNull/TcgPpVendorLibNull.inf
   SecurityPkg/Library/TcgEventLogRecordLib/TcgEventLogRecordLib.inf
+  SecurityPkg/Library/Tcg2InitEventLib/Tcg2InitEventLib.inf ## MU_CHANGE
```
spbrogan, line 275: why is this only for ia?

spbrogan (changes requested):

SecurityPkg/Include/Library/Tcg2InitEventLib.h
makubacki: At least EV_EFI_HCRTM_EVENT is defined in the TCG PFP Spec.

makubacki: I was also considering feedback that this should either be a more generic "event" library to justify the library class or moved somewhere else.

PR #407, opened 2025-05-02
[CHERRY-PICK] SecurityPkg: Prevent invalid DBX from being set

Flickdm:
# Description

This PR prevents setting the Dbx variable if the Default being provided is less than the size of the EFI_SIGNATURE_LIST structure. This is to prevent the setting of an invalid Dbx value which would cause the system to fail to boot.

Additionally, this can be used to signal that setting the DBX should leave DBX undefined for Platforms that want to let the OS be the sole servicer of the DBX.

Breakdown of the math is as follows:

1. **`sizeof(EFI_SIGNATURE_LIST)`**:
   - This is the size of the `EFI_SIGNATURE_LIST` structure itself, which includes: - `EFI_GUID SignatureType` (16 bytes) - `UINT32 SignatureListSize` (4 bytes) - `UINT32 SignatureHeaderSize` (4 bytes) - `UINT32 SignatureSize` (4 bytes)
   - Total: `16 + 4 + 4 + 4 = 28 bytes`

2. **`SignatureHeaderSize`**:
   - This is the size of the optional signature header. If no header is provided, this value is `0`.

3. **`SignatureSize`**:
   - This is the size of each `EFI_SIGNATURE_DATA` entry. For an empty list, this value is `0`.

The total size of an empty `EFI_SIGNATURE_LIST` is:
```c
sizeof(EFI_SIGNATURE_LIST) + SignatureHeaderSize
```

1. **No Signature Header**:
   - If `SignatureHeaderSize = 0`, the size is: ```c 28 + 0 = 28 bytes ```

2. **With a Signature Header**:
   - If `SignatureHeaderSize = 16` (example size for a header), the size is: ```28 + 16 = 44 bytes ```

- **Minimum Size**: `28 bytes` (if `SignatureHeaderSize = 0`).
- **Additional Size**: Add the value of `SignatureHeaderSize` if a header is included.

- [ ] Breaking change?
- [X] Impacts security?
  - Allows platforms to leave the DBX undefined. While this isnt a security concern if the OS services the DBX this should be understood by platform integrators.
- [ ] Includes tests?

## How This Was Tested

On a physical platform using defaults from https://github.com/microsoft/secureboot_objects/pull/174

## Integration Instructions

N/A

### DIFF
--- a/SecurityPkg/Library/SecureBootVariableLib/SecureBootVariableLib.c
+++ b/SecurityPkg/Library/SecureBootVariableLib/SecureBootVariableLib.c
@@ -23,6 +23,21 @@
 #include <Library/SecureBootVariableLib.h>
 #include <Library/PlatformPKProtectionLib.h>
 
+//
+// This is the minimum size an EFI_SIGNATURE_LIST and data can be considered valid.
+//
+// The UEFI specification defines this as:
+//  Each signature list is a list of signatures of one type, identified by SignatureType.
+//  The signature list contains a header and then an array of zero or more signatures in
+//  the format specified by the header. The size of each signature in the signature list
+//  is specified by SignatureSize.
+//
+// This is the size of a signature list with some data in it. Effectively 44 bytes are
+// required to set the DBX. However in reality, the size of a real payload will be larger
+// since this miniumum size does not consider the SIGNATURE_LIST
+//
+#define MINIMUM_VALID_SIGNATURE_LIST  (sizeof(EFI_SIGNATURE_LIST) + sizeof(EFI_SIGNATURE_DATA) - 1)
+
 // This time can be used when deleting variables, as it should be greater than any variable time.
 EFI_TIME  mMaxTimestamp = {
   0xFFFF,     // Year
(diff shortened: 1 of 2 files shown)


apop5 (approved):

makubacki:
Can you please clean up the first paragraph of the PR description?

> This PR prevents setting the Dbx variable the Default being provided is less than the size of the EFI_SIGNATURE_LIST structure. This is to prevent the setting of ab invalid DBX which would cause the system to fail to boot.

I think it is meant to say:

> This PR prevents setting the Dbx variable if the Default being provided is less than the size of the EFI_SIGNATURE_LIST structure. This is to prevent the setting of an invalid Dbx value which would cause the system to fail to boot.

SecurityPkg/Library/SecureBootVariableLib/SecureBootVariableLib.c
```diff
@@ -23,6 +23,21 @@
 #include <Library/SecureBootVariableLib.h>
 #include <Library/PlatformPKProtectionLib.h>
 
+//
+// This is the minimum size of a EFI_SIGNATURE_LIST and data and still be valid.
```
makubacki:
```suggestion
// This is the minimum size an EFI_SIGNATURE_LIST and data can be considered valid.
```

Javagedes (approved):

PR #423, opened 2025-06-27
[CHERRY-PICK] Move to TPM over FFA support from EDK2 [Rebase & FF]

kuqin12:
## Description

This change reverts the original implementation of TPM over FF-A in EDK2 and moves it to the EDK2 implementation.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested on QEMU SBSA and booted to UEFI shell.

## Integration Instructions

The `Tpm2DeviceLibFfa` now resides in `SecurityPkg`


apop5 (approved):

Raymond-MS (approved):
LGTM! - Just want to comment that I have a minor concern that the TPM library may need to change based on the updates to the service where it will yield for long running commands. From what I could tell I don't think we handle that situation here in our current implementation.

makubacki (approved):

kuqin12:
> LGTM! - Just want to comment that I have a minor concern that the TPM library may need to change based on the updates to the service where it will yield for long running commands. From what I could tell I don't think we handle that situation here in our current implementation.

Yeah, agreed. When that happens, we might want to do it starting from FFA lib. I was looking at it a while back and the FFA lib will need quite some change to cover the yield case.

PR #428, opened 2025-07-21
[REBASE & FF] [CHERRY-PICK] SecurityPkg: Add Tpm2DumpLib

os-d:
## Description

This PR adds Tpm2DumpLib to SecurityPkg. It is used to share code between different Tpm2DeviceLibs that need to dump out common TPM transactions for debugging purposes.

This is not a revamp of Tpm2DebugLib, which was rejected. Tpm2DebugLib was attempting to control debug prints by switching out libraries classes. This is not the purpose of Tpm2DumpLib. This new library is only attempting to share common code between different modules. Control of the debug prints is left to the platform on whether they set DEBUG_SECURITY level or not.

This is cherry-picked from an open edk2 PR to unblock Mu work: https://github.com/tianocore/edk2/pull/11320. It is being cherry-picked while the edk2 PR is in flight because it is blocking PRs in [mu_feature_ffa](https://github.com/microsoft/mu_feature_ffa/pull/33), which depends on logic that will be moved to Tpm2DumpLib. It was previously in Tpm2DebugLib, which was dropped from edk2/mu_tiano_plus.

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Just logging changes.

## Integration Instructions

This is marked as a breaking change because a new library class is added. Platforms must add Tpm2DumpLib to their DSCs if they depend on Tpm2DeviceLibDTpm or Tpm2DeviceLibFfa.


kuqin12 (approved):

apop5 (approved):

makubacki (approved):

SecurityPkg/Include/Library/Tpm2DumpLib.h
makubacki, line 1: Is this really needed before the edk2 PR completes? It already has 5 commits and then the usual practice of reverting and cherry-picking again thrashes the branch more (and it doesn't have MU_CHANGE tags in the meantime). Just wanted to make sure that's necessary as this doesn't seem like an essential change by itself. Or, if it is for an integration reason, can you put the details in the PR description?

os-d: @kuqin12 asked me to put it up :). Kun, can you share the integration reason?

kuqin12: This is because the other PRs that needs to use the original interface is breaking PR pipeline: https://github.com/microsoft/mu_feature_ffa/pull/33.

Depending on how long the edk2 PR will take :) I do not mind holding off the ffa change for another week. But if it will be another month (which normally how long my PR will take), this hassle still feels worthy.

makubacki: @os-d, can you connect the dots in the PR description to the break @kuqin12 mentioned and how this mitigates that?

Given this is only targeting the dev branch, the thrash is less of a concern in the mu_tiano_plus branch itself. More so that it's a breaking change that requires consumers to take action and if edk2 requests changes that necessitate further integration in the repos that use this intermediate code, the thrash has a ripple effect on those consumers.

makubacki: That said, as far as this conversation thread goes, I just wanted to get wider context added to the PR description and raise the general concern. Please close and proceed as needed.

os-d: I'll update the PR description, but @kuqin12 for that, it probably is just simpler to comment out the Dump[Input|Output]Block calls in that lib. There is just one of each and it is just code that prints. From activity on the edk2 PR last night, I'm not sure this is going to get accepted as is, so it will be more churn, probably and we'll have to update that code anyway.

But, I've updated the PR description and you can decide what you want to do.

os-d:
Discussed offline with @kuqin12 . Closing this PR to avoid the churn that edk2 changes will have and mu_feature_ffa will just comment out the two calls for now.

PR #431, opened 2025-07-30
Increase TcgLog Area length from 64KB to 128KB

Flickdm:
## Description

It has been seen on some code paths that the TCG Log size is not large enough. This problem is likely to only get worse due to the increasing size of the DBX. This is a preventative measure to ensure platforms continue to boot.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- [ ] TODO - Test on real platform

## Integration Instructions

N/A

### DIFF
--- a/SecurityPkg/SecurityPkg.dec
+++ b/SecurityPkg/SecurityPkg.dec
@@ -556,7 +556,9 @@
   ## This PCD defines minimum length(in bytes) of the system preboot TCG event log area(LAML).
   #  For PC Client Implementation spec up to and including 1.2 the minimum log size is 64KB.
   # @Prompt Minimum length(in bytes) of the system preboot TCG event log area(LAML).
-  gEfiSecurityPkgTokenSpaceGuid.PcdTcgLogAreaMinLen|0x10000|UINT32|0x00010017
+  # MU_CHANGE [BEGIN]
+  gEfiSecurityPkgTokenSpaceGuid.PcdTcgLogAreaMinLen|0x20000|UINT32|0x00010017
+  # MU_CHANGE [END]
 
   ## This PCD defines length(in bytes) of the TCG2 Final event log area.
   # @Prompt Length(in bytes) of the TCG2 Final event log area.


Flickdm:
Do not merge until testing has been performed. This message will be deleted.

kuqin12 (approved):

cfernald (approved):

makubacki:
Are you planning to change this in Tianocore?

Flickdm:
> Are you planning to change this in Tianocore?

That's a good point and worth doing

The background here is that there are likely two issues at play here.

1. The DBX alone could grow to 64KB thus we have a scenario in in market devices where naturally they will fill up and this is acting as just more room to grow.
    - Likely enough of a reason to update this on EDK2  
2. On the frontpage - there appears some behavior that causes the secure boot variables to be entered twice into the TCGLog and when the DBX and needs to be investigated further.
   - Could impact EDK2 but will require investigation

Flickdm:
> > Are you planning to change this in Tianocore?
> 
> That's a good point and worth doing
> 
> The background here is that there are likely two issues at play here.
> 
> 1. The DBX alone could grow to 64KB thus we have a scenario in in market devices where naturally they will fill up and this is acting as just more room to grow.
>    
>    * Likely enough of a reason to update this on EDK2
> 2. On the frontpage - there appears some behavior that causes the secure boot variables to be entered twice into the TCGLog and when the DBX and needs to be investigated further.
>    
>    * Could impact EDK2 but will require investigation

Talking it over with @kuqin12 

We think this is likely to run into a lot of pain during upstreaming and the potentially better path would be to upstream the assert that caught this bug to begin with https://github.com/microsoft/mu_tiano_plus/blob/80867d4a30b4368b6c2f6871b0d8a0fa4ff28c98/SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c#L879

makubacki:
> Talking it over with @kuqin12
> 
> We think this is likely to run into a lot of pain during upstreaming and the potentially better path would be to upstream the assert that caught this bug to begin with
> 
> https://github.com/microsoft/mu_tiano_plus/blob/80867d4a30b4368b6c2f6871b0d8a0fa4ff28c98/SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c#L879

My main concern is having an override in SecurityPkg.dec that does not have an upstream path.

If that assert is done, I think we should change this in platform DSC files like everyone else.

PR #450, opened 2025-09-11
FdtLib: fdt_rw.c - fix undefined pointer overflow behavior

antklein:
## Description

This change implements a fix for CodeQL issue: https://codeql.github.com/codeql-query-help/cpp/cpp-pointer-overflow-check/
It removes the potential for pointer overflow behavior.

This PR may be ignored as the edk2 stable tag for 202508 deletes the entire FdtLib anyways.
EDK2 PR that was rejected in favor of removing support of FdtLib. https://github.com/tianocore/edk2/pull/10955

- [ ] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

* I validated the CI build tests continue to pass. I also validated the change via QEMU emulation.

## Integration Instructions

* N/A

### DIFF
--- a/EmbeddedPkg/Library/FdtLib/fdt_rw.c
+++ b/EmbeddedPkg/Library/FdtLib/fdt_rw.c
@@ -124,7 +124,7 @@ _fdt_splice (
   char  *p   = splicepoint;
   char  *end = (char *)fdt + _fdt_data_size (fdt);
 
-  if (((p + oldlen) < p) || ((p + oldlen) > end)) {
+  if ((unsigned int)oldlen > (end - p)) {
     return -FDT_ERR_BADOFFSET;
   }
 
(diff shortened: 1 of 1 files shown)


makubacki:
@antklein, given this is going away anyway and the maintainer suggested not to use it, should we also just drop it and only support BaseFdtLib in MdePkg?

antklein:
> @antklein, given this is going away anyway and the maintainer suggested not to use it, should we also just drop it and only support BaseFdtLib in MdePkg?

I agree that we shouldn't make this change and just wait for 202508, which drops FdtLib altogether. I don't know how FdtLib compares to BaseFdtLib in the 202502 branch. If we can move, we should try.

makubacki:
> > @antklein, given this is going away anyway and the maintainer suggested not to use it, should we also just drop it and only support BaseFdtLib in MdePkg?
> 
> I agree that we shouldn't make this change and just wait for 202508, which drops FdtLib altogether. I don't know how FdtLib compares to BaseFdtLib in the 202502 branch. If we can move, we should try.

Sounds good. If cherry-picks are needed to use BaseFdtLib in 202502, we can do that. Based on this conversation, I'll close this PR.

PR #482, opened 2025-12-04
[FF-A] Update TPM Service for Manage Locality

Raymond-MS:
## Description

Added support for the manage locality command in the TPM service header.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested with QEMU SBSA with TPM enabled. Ran with FfaPartitionTestApp.

## Integration Instructions

N/A


SecurityPkg/Include/Guid/Tpm2ServiceFfa.h
```diff
@@ -21,6 +21,7 @@
 #define TPM2_FFA_REGISTER_FOR_NOTIFICATION     0x0f000301
 #define TPM2_FFA_UNREGISTER_FROM_NOTIFICATION  0x0f000401
 #define TPM2_FFA_FINISH_NOTIFIED               0x0f000501
+#define TPM2_FFA_MANAGE_LOCALITY               0x1f000001 // MU_CHANGE
```
kuqin12: Is this spec defined? If so, can you add the spec reference?

Raymond-MS: I don't believe it was officially added to the spec for FF-A over CRB. This is something that Stuart was going to work on and I haven't heard anything back from when I approved of the change.

kuqin12: Okay, I think we need to at least add the _targeting spec number_ before adding this to the code.

kuqin12 (approved):

cfernald (approved):

PR #489, opened 2026-01-13
[CHERRY-PICK] SecurityPkg/Library: use ArmFfaGetPartitionInfo() in Tpm2ServiceFfaLib

kuqin12:
## Description

This patch adds ArmFfaLibPartitionInfoGetRegs() in ArmFfaLib
As ArmFfaLibPartitionInfoGetRegs() is added, Normal world ArmFfLib Rx/Tx buffer is not a madatory.
That's why return EFI_UNSUPPORTED while Rx/Tx buffer mapping in ArmFfaLib constructor
can be consider as valid return If ARM_FFA_PARTITION_INFO_GET_REGS is supported by SPMC.

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This is tested on physical ARM64 platform.

## Integration Instructions

Need latest `MdeModulePkg` that contains the interface change.

### DIFF
--- a/SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2ServiceFfaRaw.c
+++ b/SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2ServiceFfaRaw.c
@@ -101,71 +101,27 @@ FfaTpm2GetServicePartitionId (
   )
 {
   EFI_STATUS              Status;
-  UINT32                  Count;
-  UINT32                  Size;
-  EFI_FFA_PART_INFO_DESC  *TpmPartInfo;
-  VOID                    *TxBuffer;
-  UINT64                  TxBufferSize;
-  VOID                    *RxBuffer;
-  UINT64                  RxBufferSize;
-  UINT16                  PartId;
+  EFI_FFA_PART_INFO_DESC  TpmPartInfo;
 
   if (PartitionId == NULL) {
-    Status = EFI_INVALID_PARAMETER;
-    goto Exit;
+    return EFI_INVALID_PARAMETER;
   }
 
-  Status = ArmFfaLibPartitionIdGet (&PartId);
-  if (EFI_ERROR (Status)) {
-    DEBUG ((
-      DEBUG_ERROR,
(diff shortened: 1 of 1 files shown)


apop5 (approved):

SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2ServiceFfaRaw.c
```diff
@@ -101,71 +101,27 @@ FfaTpm2GetServicePartitionId (
...
-  } else {
-    TpmPartInfo  = (EFI_FFA_PART_INFO_DESC *)RxBuffer;
-    *PartitionId = TpmPartInfo->PartitionId;
-    if (TpmPartInfo->PartitionId == TPM2_FFA_PARTITION_ID_INVALID) {
-      /*
-       * Tpm partition id never be TPM2_FFA_PARTITION_ID_INVALID.
-       */
-      Status = EFI_DEVICE_ERROR;
-    }
+  *PartitionId = TpmPartInfo.PartitionId;
```
Raymond-MS, line 116: Nit: Do we want to populate the PartitionId before validating it? I don't think it matters if we are returning a device error but we could just move this to after the check to avoid it altogether.

Raymond-MS (approved):

cfernald (approved):

SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2ServiceFfaRaw.c
kuqin12: It would be better that way. But this is a cherry-pick and it is probably too nuanced to bring it to edk2. Will resolve for now. But let me know if you think otherwise.

Raymond-MS: I have no strong feelings, no worries.

PR #492, opened 2026-01-22
[FF-A][TPM] Implement Tpm2StartupLib

Raymond-MS:
## Description

Implemented the Tpm2StartupLib which is meant to handle TPM initialization as well as any pre-DXE measurements. Updated the TPM libraries to include SEC versions which do not have global variables. This was causing issues with boot on our PEI-less QEMU SBSA platform when attempting to init the TPM in the SEC phase. A NULL version is also added for platforms who do not need any TPM initialization. The default is the NULL instance.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Built and ran with TPM enabled on our QEMU SBSA build. Confirmed TPM communication and boot to UEFI shell.

## Integration Instructions

If needed, link in the non-NULL instance of the library into your .dsc and call the init function in the proper phase before DXE is loaded.

### DIFF
--- a/SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2DeviceSecLibFfa.inf
+++ b/SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2DeviceSecLibFfa.inf
@@ -26,7 +26,7 @@
 [Sources.common]
   Tpm2DeviceLibFfa.c
   Tpm2ServiceFfaRaw.c
-  Tpm2DeviceLibFfaBase.c
+  Tpm2DeviceSecLibFfaBase.c ## MU_CHANGE
   Tpm2Ptp.c
   Tpm2DeviceLibFfa.h
   Tpm2InfoSecFfa.c
--- a/SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2DeviceSecLibFfaBase.c
+++ b/SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2DeviceSecLibFfaBase.c
@@ -14,8 +14,6 @@
 #include <Library/Tpm2DeviceLib.h>
 #include "Tpm2DeviceLibFfa.h"
 
-UINT8  mCRBIdleByPass;
-
 /**
   Return cached PTP CRB interface IdleByPass state.
 
@@ -26,7 +24,7 @@ GetCachedIdleByPass (
   VOID
   )
 {
-  return mCRBIdleByPass;
+  return Tpm2GetIdleByPass ((VOID *)(UINTN)PcdGet64 (PcdTpmBaseAddress));
 }
 
 /**
@@ -42,20 +40,9 @@ InternalTpm2DeviceLibFfaConstructor (
   VOID
   )
 {
-  EFI_STATUS  Status;
-
-  mCRBIdleByPass = 0xFF;
-
   if (PcdGet64 (PcdTpmBaseAddress) == 0) {
     return EFI_NO_MAPPING;
   }
 
-  Status = ValidateTpmInterfaceType ();
-  if (EFI_ERROR (Status)) {
-    return Status;
-  }
-
-  mCRBIdleByPass = Tpm2GetIdleByPass ((VOID *)(UINTN)PcdGet64 (PcdTpmBaseAddress));
-
-  return EFI_SUCCESS;
+  return ValidateTpmInterfaceType ();
 }
--- a/SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2ServiceFfaRaw.c
+++ b/SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2ServiceFfaRaw.c
@@ -26,18 +26,16 @@
 
 #include "Tpm2DeviceLibFfa.h"
 
-UINT16  mFfaTpm2PartitionId = TPM2_FFA_PARTITION_ID_INVALID;
-
 /**
   Check the return status from the FF-A call and returns EFI_STATUS
 
-  @param EFI_LOAD_ERROR  FF-A status code returned in x0
+  @param TpmReturnStatus  FF-A status code returned in x0
 
   @retval EFI_SUCCESS    The entry point is executed successfully.
 **/
 EFI_STATUS
 TranslateTpmReturnStatus (
-  UINTN  TpmReturnStatus
+  IN UINTN  TpmReturnStatus
   )
 {
   EFI_STATUS  Status;
@@ -142,27 +140,25 @@ Tpm2GetInterfaceVersion (
 {
   EFI_STATUS       Status;
   DIRECT_MSG_ARGS  FfaDirectReq2Args;
+  UINT16           FfaTpm2PartitionId;
 
   if (Version == NULL) {
-    Status = EFI_INVALID_PARAMETER;
-    goto Exit;
+    return EFI_INVALID_PARAMETER;
   }
 
-  if (mFfaTpm2PartitionId == TPM2_FFA_PARTITION_ID_INVALID) {
-    GetTpmServicePartitionId (&mFfaTpm2PartitionId);
-  }
+  GetTpmServicePartitionId (&FfaTpm2PartitionId);
 
   ZeroMem (&FfaDirectReq2Args, sizeof (DIRECT_MSG_ARGS));
   FfaDirectReq2Args.Arg0 = TPM2_FFA_GET_INTERFACE_VERSION;
 
-  Status = ArmFfaLibMsgSendDirectReq2 (mFfaTpm2PartitionId, &gTpm2ServiceFfaGuid, &FfaDirectReq2Args);
+  Status = ArmFfaLibMsgSendDirectReq2 (FfaTpm2PartitionId, &gTpm2ServiceFfaGuid, &FfaDirectReq2Args);
   while (Status == EFI_INTERRUPT_PENDING) {
     // We are assuming vCPU0 of the TPM SP since it is UP.
-    Status = ArmFfaLibRun (mFfaTpm2PartitionId, 0x00, &FfaDirectReq2Args);
+    Status = ArmFfaLibRun (FfaTpm2PartitionId, 0x00, &FfaDirectReq2Args);
   }
 
   if (EFI_ERROR (Status)) {
-    goto Exit;
+    return Status;
   }
 
   Status = TranslateTpmReturnStatus (FfaDirectReq2Args.Arg0);
@@ -171,7 +167,6 @@ Tpm2GetInterfaceVersion (
     *Version = FfaDirectReq2Args.Arg1;
   }
 
-Exit:
   return Status;
 }
 
@@ -193,34 +188,29 @@ Tpm2GetFeatureInfo (
 {
   EFI_STATUS       Status;
   DIRECT_MSG_ARGS  FfaDirectReq2Args;
+  UINT16           FfaTpm2PartitionId;
 
   if (FeatureInfo == NULL) {
-    Status = EFI_INVALID_PARAMETER;
-    goto Exit;
+    return EFI_INVALID_PARAMETER;
   }
 
-  if (mFfaTpm2PartitionId == TPM2_FFA_PARTITION_ID_INVALID) {
-    GetTpmServicePartitionId (&mFfaTpm2PartitionId);
-  }
+  GetTpmServicePartitionId (&FfaTpm2PartitionId);
 
   ZeroMem (&FfaDirectReq2Args, sizeof (DIRECT_MSG_ARGS));
   FfaDirectReq2Args.Arg0 = TPM2_FFA_GET_FEATURE_INFO;
   FfaDirectReq2Args.Arg1 = TPM_SERVICE_FEATURE_SUPPORT_NOTIFICATION;
 
-  Status = ArmFfaLibMsgSendDirectReq2 (mFfaTpm2PartitionId, &gTpm2ServiceFfaGuid, &FfaDirectReq2Args);
+  Status = ArmFfaLibMsgSendDirectReq2 (FfaTpm2PartitionId, &gTpm2ServiceFfaGuid, &FfaDirectReq2Args);
   while (Status == EFI_INTERRUPT_PENDING) {
     // We are assuming vCPU0 of the TPM SP since it is UP.
-    Status = ArmFfaLibRun (mFfaTpm2PartitionId, 0x00, &FfaDirectReq2Args);
+    Status = ArmFfaLibRun (FfaTpm2PartitionId, 0x00, &FfaDirectReq2Args);
   }
 
   if (EFI_ERROR (Status)) {
-    goto Exit;
+    return Status;
   }
 
-  Status = TranslateTpmReturnStatus (FfaDirectReq2Args.Arg0);
-
-Exit:
-  return Status;
+  return TranslateTpmReturnStatus (FfaDirectReq2Args.Arg0);
 }
 
 /**
@@ -241,30 +231,26 @@ Tpm2ServiceStart (
 {
   EFI_STATUS       Status;
   DIRECT_MSG_ARGS  FfaDirectReq2Args;
+  UINT16           FfaTpm2PartitionId;
 
-  if (mFfaTpm2PartitionId == TPM2_FFA_PARTITION_ID_INVALID) {
-    GetTpmServicePartitionId (&mFfaTpm2PartitionId);
-  }
+  GetTpmServicePartitionId (&FfaTpm2PartitionId);
 
   ZeroMem (&FfaDirectReq2Args, sizeof (DIRECT_MSG_ARGS));
   FfaDirectReq2Args.Arg0 = TPM2_FFA_START;
   FfaDirectReq2Args.Arg1 = (FuncQualifier & 0xFF);
   FfaDirectReq2Args.Arg2 = (LocalityQualifier & 0xFF);
 
-  Status = ArmFfaLibMsgSendDirectReq2 (mFfaTpm2PartitionId, &gTpm2ServiceFfaGuid, &FfaDirectReq2Args);
+  Status = ArmFfaLibMsgSendDirectReq2 (FfaTpm2PartitionId, &gTpm2ServiceFfaGuid, &FfaDirectReq2Args);
   while (Status == EFI_INTERRUPT_PENDING) {
     // We are assuming vCPU0 of the TPM SP since it is UP.
-    Status = ArmFfaLibRun (mFfaTpm2PartitionId, 0x00, &FfaDirectReq2Args);
+    Status = ArmFfaLibRun (FfaTpm2PartitionId, 0x00, &FfaDirectReq2Args);
   }
 
   if (EFI_ERROR (Status)) {
-    goto Exit;
+    return Status;
   }
 
-  Status = TranslateTpmReturnStatus (FfaDirectReq2Args.Arg0);
-
-Exit:
-  return Status;
+  return TranslateTpmReturnStatus (FfaDirectReq2Args.Arg0);
 }
 
 /**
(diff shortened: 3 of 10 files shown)


SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2DeviceSecLibFfa.inf
```diff
@@ -25,8 +25,8 @@
 
 [Sources.common]
   Tpm2DeviceLibFfa.c
-  Tpm2ServiceFfaRaw.c
```
Raymond-MS, line 28: Need to add MU_CHANGE tags

SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2DeviceSecLibFfaBase.c
```diff
@@ -0,0 +1,66 @@
...
+InternalTpm2DeviceLibFfaConstructor (
+  VOID
+  )
+{
+  EFI_STATUS  Status;
+
+  if (PcdGet64 (PcdTpmBaseAddress) == 0) {
+    Status = EFI_NO_MAPPING;
+    goto Exit;
+  }
```
makubacki, line 45: For simple cases like this, it's generally preferred to just return the status code. The `goto` approach is more useful in cases where the `Exit` label has cleanup logic.

SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2DeviceSecLibFfaBase.c
```diff
@@ -0,0 +1,66 @@
...
+#include <IndustryStandard/ArmFfaSvc.h>
+#include <Library/BaseLib.h>
+#include <Library/BaseMemoryLib.h>
+#include <Library/DebugLib.h>
+#include <Library/Tpm2DeviceLib.h>
+#include <IndustryStandard/ArmStdSmc.h>
+#include <IndustryStandard/Tpm20.h>
+#include <Library/TimerLib.h>
+
+#include "Tpm2DeviceLibFfa.h"
```
makubacki: Are all of these includes really needed? I didn't look up the references but it seems like a lot for the relatively little content in this file.

SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2DeviceSecLibFfaBase.c
```diff
@@ -0,0 +1,66 @@
...
+
+  Status = ValidateTpmInterfaceType ();
+  if (EFI_ERROR (Status)) {
+    goto Exit;
+  }
+
+  Status = EFI_SUCCESS;
+
+Exit:
+  return Status;
```
makubacki:
```suggestion
  return ValidateTpmInterfaceType ();
```

SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2DeviceSecLibFfaBase.c
makubacki:
```suggestion
  if (PcdGet64 (PcdTpmBaseAddress) == 0) {
    return EFI_NO_MAPPING;
  }
```

SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2ServiceSecFfaRaw.c
```diff
@@ -0,0 +1,423 @@
...
+#include <Library/DebugLib.h>
+#include <Library/Tpm2DeviceLib.h>
+#include <Library/ArmFfaLib.h>
+
+#include "Tpm2DeviceLibFfa.h"
+
+/**
+  Check the return status from the FF-A call and returns EFI_STATUS
+
+  @param EFI_LOAD_ERROR  FF-A status code returned in x0
```
makubacki:
```suggestion
  @param TpmReturnStatus  FF-A status code returned in x0
```

SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2ServiceSecFfaRaw.c
```diff
@@ -0,0 +1,423 @@
...
+
+  @retval EFI_SUCCESS    The entry point is executed successfully.
+**/
+EFI_STATUS
+TranslateTpmReturnStatus (
+  UINTN  TpmReturnStatus
```
makubacki:
```suggestion
  IN UINTN  TpmReturnStatus
```

SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2ServiceSecFfaRaw.c
```diff
@@ -0,0 +1,423 @@
...
+  VOID                    *TxBuffer;
+  UINT64                  TxBufferSize;
+  VOID                    *RxBuffer;
+  UINT64                  RxBufferSize;
+  UINT16                  PartId;
+
+  if (PartitionId == NULL) {
+    Status = EFI_INVALID_PARAMETER;
+    goto Exit;
+  }
```
makubacki: Convention is to `return EFI_INVALID_PARAMETER` in a case like this.

SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2ServiceSecFfaRaw.c
```diff
@@ -0,0 +1,423 @@
...
+  Status = ArmFfaLibPartitionInfoGet (
+             &gTpm2ServiceFfaGuid,
+             FFA_PART_INFO_FLAG_TYPE_DESC,
+             &Count,
+             &Size
+             );
+  if (EFI_ERROR (Status)) {
+    DEBUG ((DEBUG_ERROR, "Failed to get Tpm2 partition info. Status: %r\n", Status));
+    goto RxRelease;
+  }
```
makubacki: If this is the only place that goes to the `RxRelease` label, convention is to just call `ArmFfaLibRxRelease (PartId)` followed by returning the status code.

SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2ServiceSecFfaRaw.c
makubacki: I won't leave this on all of the functions but in similar cases, they can be simplified too.

SecurityPkg/Library/Tpm2StartupLib/Tpm2StartupLib.inf
```diff
@@ -0,0 +1,45 @@
+## @file
+#  Handles TPM 2.0 Startup/Initialization.
+#
+#  Includes setup of initial TPM state as well as any pre-DXE
+#  measurements. Also handles any pre-DXE logging by generating
+#  HOBs.
+#
+# Copyright (c) 2026, Microsoft Corporation.  All rights reserved. <BR>
```
makubacki:
```suggestion
# Copyright (c) Microsoft Corporation.
```

makubacki:
Fyi, my comments were quick and superficial things I noticed when the PR came through. Will let the other assigned reviewers review in more detail.

SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2DeviceSecLibFfaBase.c
Raymond-MS: Agreed!

SecurityPkg/Library/Tpm2StartupLibNull/Tpm2StartupLibNull.inf
```diff
@@ -0,0 +1,32 @@
...
+
+[Sources]
+  Tpm2StartupLibNull.c
+
+[Packages]
+  MdePkg/MdePkg.dec
+  SecurityPkg/SecurityPkg.dec
+
+[LibraryClasses]
+  BaseLib
```
cfernald, line 31: Doesn't actually use any libraries, right?

SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2ServiceSecFfaRaw.c
```diff
@@ -0,0 +1,361 @@
...
+  )
+{
+  EFI_STATUS              Status;
+  EFI_FFA_PART_INFO_DESC  TpmPartInfo;
+
+  if (PartitionId == NULL) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  Status = ArmFfaLibGetPartitionInfo (&gTpm2ServiceFfaGuid, &TpmPartInfo);
```
cfernald: It seems most the code is identical with the exception of the ArmFfaLibGetPartitionInfo versus stashed value. Did you consider a common implementation and creating a "GetTpmPartitionId" function that is forked to either grab the global or query?

SecurityPkg/Library/Tpm2StartupLibNull/Tpm2StartupLibNull.inf
Raymond-MS: At the moment, no, I don't think so. However, I was going based on other NULL based implementations. I figured having these two libraries as a base would be good in the off chance we wanted to add some debug statements, etc.

SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2ServiceSecFfaRaw.c
Raymond-MS: This is how we are handling the globals in ArmFfaCommon, however, from my slight bit if research, there are no phase specific files for us to implement a function to do this. We would need to add one which I think is more work than having an SEC version of this file.

cfernald: Sure, but instead of duplicating and slightly changing the code, I would posit it would be better to make the existing Tpm2ServicesFfaRaw file the "common" by calling a new function to be implemented in Tpm2Device*LibFfaBase or another file that is forked for SEC and everything else.

An alternative could be just always call `ArmFfaLibPartitionIdGet` and have the FFA lib cache that information instead, to prevent having to cache at all the different possible consumers. 

My complaint is just this file is mostly duplicate code

Raymond-MS: Sounds good, I like that approach. Working on it now. :)

Wanted to mention that it looks as though we already had the service ID being cached in the FF-A lib through a PCD. So there was no need to cache it again as a static in that file. I removed it and it should resolve the issue.

SecurityPkg/Library/Tpm2StartupLibNull/Tpm2StartupLibNull.inf
```diff
@@ -0,0 +1,32 @@
+## @file
+#  NULL Instance of Tpm2StartupLib
+#
+# Copyright (c) 2026, Microsoft Corporation.  All rights reserved. <BR>
```
kuqin12:
```suggestion
# Copyright (c), Microsoft Corporation.
```

SecurityPkg/SecurityPkg.dec
```diff
@@ -51,6 +51,11 @@
   #
   Tpm2HelpLib|Include/Library/Tpm2HelpLib.h
 
+  ## MU_CHANGE
+  ##  @libraryclass This library provides TPM initialization and setup functionality.
+  ##
+  Tpm2StartupLib|Include/Library/Tpm2StartupLib.h
```
kuqin12, line 57: For the sake of upstreaming, we will need these change to break up into individual commits (Tpm2StartupLib, and change of tpm device lib)

SecurityPkg/Library/Tpm2StartupLibNull/Tpm2StartupLibNull.c
```diff
@@ -0,0 +1,27 @@
...
+  @retval EFI_NOT_FOUND     TPM device not found
+  @retval EFI_DEVICE_ERROR  Unexpected device error
+**/
+EFI_STATUS
+EFIAPI
+Tpm2StartupInit (
+  VOID
+  )
+{
+  return EFI_SUCCESS;
```
kuqin12: You probably want to return `EFI_UNSUPPORTED`? To prevent platform misconfiguration.

SecurityPkg/Library/Tpm2StartupLib/Tpm2StartupLib.inf
```diff
@@ -0,0 +1,45 @@
...
+[Packages]
+  MdePkg/MdePkg.dec
+  MdeModulePkg/MdeModulePkg.dec
+  SecurityPkg/SecurityPkg.dec
+
+[LibraryClasses]
+  BaseLib
+  BaseMemoryLib
+  DebugLib
+  ArmFfaLib
```
kuqin12: We cannot have this here.

SecurityPkg/Library/Tpm2StartupLib/Tpm2StartupLib.c
```diff
@@ -0,0 +1,54 @@
...
+  This function initializes the TPM if required
+
+  @retval EFI_SUCCESS       TPM successfully initialized
+  @retval EFI_UNSUPPORTED   TPM is not supported
+  @retval EFI_NOT_FOUND     TPM device not found
+  @retval EFI_DEVICE_ERROR  Unexpected device error
+**/
+EFI_STATUS
+EFIAPI
+Tpm2StartupInit (
```
kuqin12, line 34: I might have missed it, but where is the caller?

SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2ServiceFfaRaw.c
```diff
@@ -144,8 +144,7 @@ Tpm2GetInterfaceVersion (
   DIRECT_MSG_ARGS  FfaDirectReq2Args;
 
   if (Version == NULL) {
-    Status = EFI_INVALID_PARAMETER;
-    goto Exit;
```
kuqin12, line 148: Why are the changes in this function necessary?

SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2ServiceFfaRaw.c
```diff
@@ -195,8 +193,7 @@ Tpm2GetFeatureInfo (
   DIRECT_MSG_ARGS  FfaDirectReq2Args;
 
   if (FeatureInfo == NULL) {
-    Status = EFI_INVALID_PARAMETER;
```
kuqin12, line 198: Same here? And others in this file.

kuqin12 (changes requested):

SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2ServiceFfaRaw.c
Raymond-MS: Please see Michael's comments for reasoning.

SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2ServiceFfaRaw.c
Raymond-MS: Please see Michael's comments for reasoning.

SecurityPkg/Library/Tpm2StartupLib/Tpm2StartupLib.c
Raymond-MS: This is called in PeilessSec.c which is in Silicon/Arm/MU_TIANO.

SecurityPkg/Library/Tpm2StartupLib/Tpm2StartupLib.inf
Raymond-MS: I would need explanation as to why, please?

SecurityPkg/Library/Tpm2StartupLibNull/Tpm2StartupLibNull.c
Raymond-MS: I am fine with this, sure.

SecurityPkg/Library/Tpm2StartupLib/Tpm2StartupLib.inf
kuqin12: Do you call FFA interface here directly?

SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2ServiceFfaRaw.c
kuqin12: I get that for new code. But why do we change that for existing code as well?

Raymond-MS: The new code was a copy of existing code and he wanted changes made to both. In this case, the new version was removed but the old version got updated. The comments would still apply.

SecurityPkg/Library/Tpm2StartupLib/Tpm2StartupLib.inf
Raymond-MS: You are correct. I removed ArmFfaLib. Thanks!

SecurityPkg/SecurityPkg.dec
Raymond-MS: Complete!

SecurityPkg/Library/Tpm2StartupLib/Tpm2StartupLib.c
```diff
@@ -0,0 +1,54 @@
...
+  )
+{
+  EFI_STATUS  Status;
+
+  DEBUG ((DEBUG_INFO, "%a - Entry\n", __func__));
+
+  // Request use of the TPM
+  Status = Tpm2RequestUseTpm ();
+  if (EFI_ERROR (Status)) {
+    DEBUG ((DEBUG_ERROR, "TPM2 not detected!\n"));
```
cfernald: Log the error too.

SecurityPkg/Library/Tpm2StartupLib/Tpm2StartupLib.c
```diff
@@ -0,0 +1,54 @@
...
+    return Status;
+  }
+
+  // Determine if initialization is requested
+  if (PcdGet8 (PcdTpm2InitializationPolicy) == 1) {
+    // Setup/Initialize the TPM
+    Status = Tpm2Startup (TPM_SU_CLEAR);
+    if (EFI_ERROR (Status)) {
+      DEBUG ((DEBUG_ERROR, "Tpm2Startup::%a - TPM failed Startup!\n", __func__));
```
cfernald: Same here (log the error)

cfernald (approved):

SecurityPkg/Library/Tpm2StartupLib/Tpm2StartupLib.c
Raymond-MS: To confirm, your request is to generate the TPM error HOB, correct?

cfernald: No, i just mean print the Status value

Raymond-MS: As per our conversation, added!

kuqin12:
nit: Please update the commit message to the edk2 style.

kuqin12 (approved):

